Issue #40, opened 2017-04-05
The crate domain does not support local hosts file


Matrix-Zhang:
may be change to crate tokio-dns-unofficial?

zonyitoo:
`ssserver` or `sslocal`?

Matrix-Zhang:
both of them

zonyitoo:
Why would you need that?

1. If your servers' domain cannot be resolved in `sslocal`, write IP addresses in configuration files.
2. If your servers' domain cannot be resolved in `ssserver` side... is that possible?

Matrix-Zhang:
the server is my vps host which i write local hostname in /etc/hosts

zonyitoo:
Ref to https://github.com/partim/domain/issues/12 .

Matrix-Zhang:
https://github.com/partim/nsswitch

zonyitoo:
This crate is still in a very early stage.

zonyitoo:
It should supports now. But it still requires more works on Windows.

zonyitoo:
Already deprecated `domain` crate and replaced it with standard name resolving provided by libc.

PR #41, opened 2017-04-05
which resolve the issue #40

Matrix-Zhang:
use tokio-dns to connect to proxy server instead of domain


zonyitoo:
This `tokio-dns-unofficial` crates uses thread pool to resolve DNS queries. I don't think this is the optimal solution. The older version of shadowsocks-rust uses a very similar solution (put the `to_socket_addrs` in a `CpuPool`).

What you proposing is to add support of reading `hosts` files in different platforms, it could be supported separately.

BTW, is the official version of shadowsocks (Python version) supports reading `hosts` files?

zonyitoo:
Closing this because I won't merge this PR.

PR #44, opened 2017-05-21
fix clippy warning

bennyyip:


tests/udp.rs
```diff
@@ -1,4 +1,4 @@
-#![allow(dead_code, unused_imports)]
+#![allow(dead_code, unused_imports,blacklisted_name)]
```
zonyitoo: missing a space between `,` and `b`

zonyitoo: Never mind, it is just a test.

PR #81, opened 2017-11-01
Add experimental SIV support

quininer:
This PR will add experimental Aes128PmacSiv, Aes256PmacSiv support.

It has some performance loss, but will be an immune nonce reuse attack.


zonyitoo:
Looks good.
I think you should also add those new ciphers into cipher support section in README.md

PR #84, opened 2017-11-21
fix increase_nonce

bennyyip:
The old implementation dose nothing to the nonce, because `prev` is 0.
See also https://github.com/jedisct1/libsodium/blob/569778b517496861a3880e0e690973bf08a52e08/src/libsodium/sodium/utils.c#L239


zonyitoo:
Ah, that's true. Thanks!

PR #123, opened 2018-08-16
Add `set_nodelay` and `set_keepalive`

eataix:
These additions are similar to those in 271e24d, which were then
removed during the transition to futures and tokio in 5bf6560.


zonyitoo:
I don't think it is a good idea to `set_nodelay`. Because we uses streaming protocol assembling, which means that we may call `send` multiple times while assembling a single protocol header (shadowsocks' protocol header).

sorz:
+1 for TCP keepalive
This should be the correct implementation for dead connection detection.
"copy timeout" may better to be disable or have a large value (like dozens hours).

PR #142, opened 2019-01-23
Misc simplification

faern:
This started out with seeing that `Plugin.addr` seemed unused. So I removed it and wanted to send a PR. To get some feedback if it was indeed unused, or if I had missed something about it.

Then I also saw that rustfmt suggested some changes. So I applied that. And then I saw Clippy had a few things to suggest. So I went ahead and applied some of the suggestions it had.

This PR should not change any behavior or exposed API. Except the `Plugin::addr()` method which is now gone.

### DIFF
--- a/src/bin/ssurl.rs
+++ b/src/bin/ssurl.rs
@@ -15,8 +15,8 @@ use shadowsocks::{
     VERSION,
 };
 
-const BLACK: &'static str = "\x1b[40m  \x1b[0m";
-const WHITE: &'static str = "\x1b[47m  \x1b[0m";
+const BLACK: &str = "\x1b[40m  \x1b[0m";
+const WHITE: &str = "\x1b[47m  \x1b[0m";
 
 fn print_qrcode(encoded: &str) {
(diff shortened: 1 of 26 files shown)


faern:
I hope this does not come across as me trying to force a style onto the project. That is not my intention. The first part is just removing things that seemed unused. And the second part is just conforming to what the standard linter suggested. :slightly_smiling_face:

zonyitoo:
Clippy warns me every time when I open this project.

I kept ignoring it all the time because old rustc didn't come with these features.

Is it ok to abandon those old rustc versions?

faern:
You already merged the PR that upgraded the crate to Rust edition 2018. Which means this will no longer build on anything older than 1.32. So this project already abandoned older rustc versions. Which I think is fully reasonable. Basically everyone is using the latest one anyway.

And since this crate is yet not usable as a library (it soon will be) no one is depending on it as a dependency. Thus there is no compatibility to worry about really.

zonyitoo:
You're right. Merged.

PR #144, opened 2019-01-25
Move logging code into reused module

faern:
There was a lot of logging setup code that was shared between the binaries. I reduced duplication a lot by sharing the code via a module. Easier to keep in sync, is one benefit.

### DIFF
--- a/src/bin/logging/mod.rs
+++ b/src/bin/logging/mod.rs
@@ -5,7 +5,7 @@ use std::{
     io::{self, Write},
 };
 
-pub fn init(without_time: bool, debug_level: u64) {
+pub fn init(without_time: bool, debug_level: u64, bin_name: &str) {
     let mut log_builder = Builder::new();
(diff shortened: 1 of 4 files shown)


src/bin/logging/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+        2 => {
+            let log_builder = log_builder.format(move |fmt, r| log_time_module(fmt, without_time, r));
+            log_builder
+                .filter(Some("sslocal"), LevelFilter::Debug)
+                .filter(Some("shadowsocks"), LevelFilter::Debug);
+        }
+        3 => {
+            let log_builder = log_builder.format(move |fmt, r| log_time_module(fmt, without_time, r));
+            log_builder
+                .filter(Some("sslocal"), LevelFilter::Trace)
```
zonyitoo: It doesn't work if you just use `sslocal` for all the binaries.

PR #145, opened 2019-01-25
Test on windows

faern:
Would be nice to be able to run CI tests on Windows. To make sure it does not break.. Also there seems to be issues building the crate on Windows right now. So would be good to get that resolved as well. As such I have marked the Windows build as OK to fail for now.

In the process `cache: cargo` got lost. I have personally not really seen any speedup with caching. Cargo generates a huge amount of data. And the time it takes to download and unpack this on build start and then pack and upload again on a finished build is usually longer than it just takes to do a clean build.

Also, to not make the CI run forever I opted for not running all three Rust versions on all platforms. Keeping `beta` and `nightly` on Linux but only testing `stable` on the other two platforms. What do you think about that?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,22 @@
+sudo: false
 language: rust
-cache: cargo
+matrix:
+    include:
+        - os: windows
+          rust: stable
 
-rust:
-    - stable
-    - beta
-    - nightly
+        - os: osx
+          rust: stable
 
-os:
-    - linux
-    - osx
+        - os: linux
+          rust: stable
+        - os: linux
+          rust: beta
+        - os: linux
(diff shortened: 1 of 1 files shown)


faern:
The separate `cargo build` step can be useful to have at times. It does not cost much extra time since the artifacts are reused from the `build` to the `test`. And the benefit is that it catches any compile errors that might happen only on a normal build due to conditional compilation for tests etc.

faern:
Ok... Apparently the Windows tests worked when run on Travis on this account. They failed when I ran them on the organization of my fork. Anyway, so removing it from the `allowed_failures` section again and back to the regular tests.

PR #165, opened 2019-08-31
Migrate to std::future, async/await

zonyitoo:
- [x] Socks5 Protocol Library
- [x] TCP Relay Tools
- [ ] DNS Resolving (dep: https://github.com/bluejekyll/trust-dns/pull/849)
    - [ ] Build with tokio's 0.2.x
- [x] `ring` upgrades to `v0.16.x`
    - [x] Ring removed support of SHA1 algorithm, uses separated crate `hkdf` and `sha1`
- [x] TCP Relay
- [x] UDP Relay
- [ ] <del>DNS Proxy</del>
- [x] Binaries (`sslocal`, `ssserver`, <del>`ssdns`</del>)
- [x] Tests
- [x] Windows

PR #166, opened 2019-09-14
add dyn keyword to disable warning when compiling

18o:


zonyitoo:
These will be added after #165 is finished.

zonyitoo:
#165 is merged.

PR #175, opened 2019-11-30
Add a tunnel app

cg31:
It can be used to benchmark.

Usage:
sstunnel -v -s 127.0.0.1:8488 -m plain -k password -b 127.0.0.1:1090 -f 127.0.0.1:5201

"-f xxx" : the target/forwad address xxx


zonyitoo:
I would prefer to move tunneling code out of `socks5_local.rs`. Let me do some modification.

PR #176, opened 2019-12-02
reduce loops and type conversions

driftluo:
reduce loops and type conversions


zonyitoo:
This was borrowed from libsodium: https://github.com/jedisct1/libsodium/blob/master/src/libsodium/sodium/utils.c#L294-L297 .

driftluo:
No matter where this code comes from, the behavior after the change is the same as before the change.

But from the point of view of the number of cycles:

before this change, the number of cycles depends on its length, it's fixed

after this change, the number of cycles it can reduce on some case

such as:
nonce: [0,0,0,0]

on [0,0,0,0] to [255,255,254,0], it use 0-3 cycles, only on [255,255,255,0] to [255,255,255,255] use 4 cycles

However, this is just a code-level difference, and if consider compiler optimizations, I don't know what it will become

PR #198, opened 2020-02-29
Multiple changes for shadowsocks-android

madeye:
This PR added:

1. Protect socket logic for Android.
2. Misc fixes for Android VPNService.

I think some rework and cleanup are still needed, especially for the interfaces.

### DIFF
--- a/src/relay/sys/unix/mod.rs
+++ b/src/relay/sys/unix/mod.rs
@@ -1,9 +1,15 @@
 use std::{
     io::{self, Error, ErrorKind},
     mem,
-    net::{SocketAddr, SocketAddrV4, SocketAddrV6},
+    net::{SocketAddr, SocketAddrV4, SocketAddrV6, TcpStream},
+    io::prelude::*,
+    time::Duration,
+    os::unix::io::{AsRawFd, RawFd},
+    os::unix::net::UnixStream,
 };
 
+use net2::*;
+use sendfd::*;
 use tokio::net::UdpSocket;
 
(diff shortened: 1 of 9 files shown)


src/relay/sys/unix/mod.rs
```diff
@@ -25,8 +34,82 @@ pub fn sockaddr_to_std(saddr: &libc::sockaddr_storage) -> io::Result<SocketAddr>
...
+
+    Ok(socket)
+}
+
 /// Create a `UdpSocket` binded to `addr`
 #[inline(always)]
 pub async fn create_udp_socket(addr: &SocketAddr) -> io::Result<UdpSocket> {
-    UdpSocket::bind(addr).await
+    let socket = UdpSocket::bind(addr).await?;
+    Ok(socket)
```
zonyitoo: Hmm? Why is that?

zonyitoo:
Windows build is broken. :(

src/relay/sys/unix/mod.rs
madeye: Yeah, we should revert the change here. I did some experiment here and forgot to revert...

madeye:
I fixed the build on Windows. But it looks the test still failed. Not sure why it happens.

zonyitoo:
`net2` is deprecated, use `socket2` instead.

Like this: https://github.com/shadowsocks/shadowsocks-rust/blob/master/src/relay/tcprelay/sys/unix/linux.rs#L82

zonyitoo:
Excellent. I am going to do some code cleanup later.

PR #202, opened 2020-03-11
Add DNS relay for Android

madeye:
It maybe also useful for `ssredir` and `sslocal`.

### DIFF
--- a/src/bin/local.rs
+++ b/src/bin/local.rs
@@ -9,6 +9,9 @@ use futures::future::{self, Either};
 use log::{error, info};
 use tokio::{self, runtime::Builder};
 
+#[cfg(feature = "dns-relay")]
+use std::net::SocketAddr;
+
 use shadowsocks::{
     acl::AccessControl,
     crypto::CipherType,
@@ -35,16 +38,36 @@ fn main() {
                 .help("Set the level of debug"),
         )
         .arg(
+            Arg::with_name("VPN_MODE")
+                .long("vpn")
+                .help("Enable VPN mode (only for Android)"),
+        )
+        .arg(
             Arg::with_name("STAT_PATH")
                 .long("stat-path")
                 .takes_value(true)
-                .help("Specify stat path (only for Android)"),
+                .help("Specify stat_path for traffic stat (only for Android)"),
         )
         .arg(
-            Arg::with_name("PROTECT_PATH")
-                .long("protect-path")
+            Arg::with_name("LOCAL_DNS_ADDR")
+                .long("local-dns")
                 .takes_value(true)
-                .help("Specify protect-path (only for Android)"),
+                .default_value("127.0.0.1:5353")
+                .help("Specify the address of local DNS server (only for Android)"),
+        )
+        .arg(
+            Arg::with_name("REMOTE_DNS_ADDR")
+                .long("remote-dns")
+                .takes_value(true)
+                .default_value("8.8.8.8:53")
+                .help("Specify the address of remote DNS server (only for Android)"),
+        )
+        .arg(
+            Arg::with_name("DNS_RELAY_ADDR")
+                .long("dns-realy")
+                .takes_value(true)
+                .default_value("127.0.0.1:5450")
+                .help("Specify the address of DNS relay (only for Android)"),
         )
         .arg(Arg::with_name("UDP_ONLY").short("u").help("Server mode UDP_ONLY"))
         .arg(Arg::with_name("TCP_AND_UDP").short("U").help("Server mode TCP_AND_UDP"))
@@ -183,13 +206,33 @@ fn main() {
         }
     };
 
+    #[cfg(feature = "dns-relay")]
+    {
+        if let Some(local_dns_addr) = matches.value_of("LOCAL_DNS_ADDR") {
+            let addr: SocketAddr = local_dns_addr.parse().expect("local dns address");
+            config.local_dns_addr = Some(addr);
+        }
+
+        if let Some(remote_dns_addr) = matches.value_of("REMOTE_DNS_ADDR") {
+            let addr: SocketAddr = remote_dns_addr.parse().expect("remote dns address");
+            config.remote_dns_addr = Some(addr);
+        }
+
+        if let Some(dns_relay_addr) = matches.value_of("DNS_RELAY_ADDR") {
+            let addr: SocketAddr = dns_relay_addr.parse().expect("dns relay address");
+            config.dns_relay_addr = Some(addr);
+        }
+    }
+
     if cfg!(target_os = "android") {
+        config.local_dns_path = Some("local_dns_path".to_string());
+
         if let Some(stat_path) = matches.value_of("STAT_PATH") {
             config.stat_path = Some(stat_path.to_string());
         }
 
-        if let Some(protect_path) = matches.value_of("PROTECT_PATH") {
-            config.protect_path = Some(protect_path.to_string());
+        if matches.is_present("VPN_MODE") {
+            config.protect_path = Some("protect_path".to_string());
         }
     }
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -871,13 +871,24 @@ pub struct Config {
     /// ACL configuration
     pub acl: Option<AccessControl>,
     /// Path to stat callback unix address, only for Android
-    pub stat_path: Option<String>,
-    /// Path to protect callback unix address, only for Android
-    pub protect_path: Option<String>,
     /// TCP Transparent Proxy type
     pub tcp_redir: RedirType,
     /// UDP Transparent Proxy type
     pub udp_redir: RedirType,
+    pub stat_path: Option<String>,
+    /// Path to protect callback unix address, only for Android
+    pub protect_path: Option<String>,
+    /// Path for local DNS resolver, only for Android
+    pub local_dns_path: Option<String>,
+    /// Interanl DNS's bind address
+    #[cfg(feature = "dns-relay")]
+    pub dns_relay_addr: Option<SocketAddr>,
+    /// Local DNS's address
+    #[cfg(feature = "dns-relay")]
+    pub local_dns_addr: Option<SocketAddr>,
+    /// Remote DNS's address
+    #[cfg(feature = "dns-relay")]
+    pub remote_dns_addr: Option<SocketAddr>,
 }
 
 /// Configuration parsing error kind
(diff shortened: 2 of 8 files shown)


src/context.rs
```diff
@@ -268,12 +282,39 @@ impl Context {
         }
     }
 
+    /// Add a record to the reverse lookup cache
+    #[cfg(feature = "dns-relay")]
+    pub fn add_to_reverse_lookup_cache(&self, addr: IpAddr, qname: String) {
+        let mut reverse_lookup_cache = self.reverse_lookup_cache.lock();
+        reverse_lookup_cache.insert(addr, qname);
+        ()
```
zonyitoo: unnecessary `()`

zonyitoo:
Why not just use dns parsers from trust-dns-proto? https://docs.rs/trust-dns-proto/0.19.3/trust_dns_proto/op/query/struct.Query.html

zonyitoo:
There was a `ssdns` binary in this project. But .. `sstunnel` can do exactly the same thing?

madeye:
It's actually an ACL based ChinaDNS, for Android integration.

I'm new to Rust and its libraries. It would be very nice if we can reuse current libraries or code.

zonyitoo:
You could just resolve the conficts. I can refactor it for you.

Mygod:
I am late to the party but why does this look so complicated...?

madeye:
@Mygod Yeah, I didn't take advantage of `trust-dns-proto` package here. It should help to remove most of code for DNS protocol itself.

src/relay/sys/unix/bsd_pf.rs
```diff
@@ -117,7 +124,10 @@ mod ffi {
         }
     }
 
+<<<<<<< HEAD
+=======
     // pub const PF_IN: libc::c_int = 1;
+>>>>>>> upstream/master
```
Mygod: It seems like there are more conflicts...

Mygod:
Also that 1000-line rust file is pretty unreadable... (I guess you copied some code from updns?)

I guess I will check back and try to learn rust by example. :)

madeye:
@Mygod Yeah, copy and paste from here: https://github.com/EmilHernvall/dnsguide

zonyitoo:
But, why not put all of theses into a separated binary `ssdns`? It seems that it is not only useful for Android, but for all platforms?

madeye:
@zonyitoo Yeah, I think we can have a `ssdns` for other platforms.

But for android, we prefer only one process, because we need the reverse lookup cache from dns-replay for domain name based ACL rules.

zonyitoo:
Still has confilcts. :(

zonyitoo:
> But for android, we prefer only one process, because we need the reverse lookup cache from dns-replay for domain name based ACL rules.

Then it is a Android only feature. I think we could just simply control it by `#[cfg(target_os = "android")]` instead of a separated feature.

madeye:
I think the `dns-relay` feature is also useful for `ssredir` (or `sstun` one day).

Actually, domain name based ACL only works if we take control of the system's DNS logic.

madeye:
The conflicts should be resolved now.

zonyitoo:
> > But for android, we prefer only one process, because we need the reverse lookup cache from dns-replay for domain name based ACL rules.
> 
> Then it is a Android only feature. I think we could just simply control it by `#[cfg(target_os = "android")]` instead of a separated feature.

How about this? @madeye

src/bin/local.rs
```diff
@@ -34,17 +37,37 @@ fn main() {
...
+                .long("remote-dns")
+                .takes_value(true)
+                .default_value("8.8.8.8:53")
+                .help("Specify the address of remote DNS server (only for Android)"),
         )
         .arg(
-            Arg::with_name("PROTECT_PATH")
-                .long("protect-path")
+            Arg::with_name("DNS_RELAY_ADDR")
+                .long("dns-realy")
```
Mygod: Typo here?

src/config.rs
```diff
@@ -871,13 +871,24 @@ pub struct Config {
     /// ACL configuration
     pub acl: Option<AccessControl>,
     /// Path to stat callback unix address, only for Android
```
Mygod: Forgot to move comment?

src/bin/local.rs
```diff
@@ -183,13 +206,33 @@ fn main() {
...
+        }
+
+        if let Some(dns_relay_addr) = matches.value_of("DNS_RELAY_ADDR") {
+            let addr: SocketAddr = dns_relay_addr.parse().expect("dns relay address");
+            config.dns_relay_addr = Some(addr);
+        }
+    }
+
     if cfg!(target_os = "android") {
+        config.local_dns_path = Some("local_dns_path".to_string());
```
Mygod: This seems unused. Maybe use this on `cfg!(target_os = "android")`? I don't see the necessity of exposing another port.

src/relay/dnsrelay/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+async fn udp_lookup(qname: &Name, qtype: RecordType, server: SocketAddr) -> Result<Message> {
+    let mut socket = UdpSocket::bind(("0.0.0.0", 0)).await?;
+
+    let mut message = Message::new();
+    let mut query = Query::new();
+
+    query.set_query_type(qtype);
+    query.set_name(qname.clone());
+
+    message.set_id(6666);
```
Mygod: Message ID should be random, no? Same for below.

src/relay/dnsrelay/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+        }
+    }
+
+    let mut message = Message::new();
+    let mut query = Query::new();
+
+    query.set_query_type(qtype);
+    query.set_name(qname.clone());
+
+    message.set_id(6666);
```
Mygod: Also here.

src/relay/dnsrelay/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+async fn socks5_lookup(qname: &Name, qtype: RecordType, socks5: SocketAddr, ns: SocketAddr) -> Result<Message> {
+    let mut stream = TcpStream::connect(socks5).await?;
+
+    // 1. Handshake
+    let hs = HandshakeRequest::new(vec![SOCKS5_AUTH_METHOD_NONE]);
+    hs.write_to(&mut stream).await?;
+    stream.flush().await?;
+
+    let hsp = HandshakeResponse::read_from(&mut stream).await?;
+    assert_eq!(hsp.chosen_method, SOCKS5_AUTH_METHOD_NONE);
```
Mygod: I think it is okay to send the entire packet at once, which is what `LocalDnsServer` does in shadowsocks-android.

src/relay/dnsrelay/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+    query.set_query_type(qtype);
+    query.set_name(qname.clone());
+
+    message.set_id(6666);
+    message.set_recursion_desired(true);
+    message.add_query(query);
+
+    let req_buffer = message.to_vec()?;
+    let size = req_buffer.len();
+    let mut size_buffer: [u8; 2] = [((size >> 8) & 0xFF) as u8, ((size >> 0) & 0xFF) as u8];
```
Mygod: https://stackoverflow.com/q/25060583/2245107

src/relay/dnsrelay/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+    );
+
+    let timeout = Some(Duration::new(5, 0));
+
+    let local_response = try_timeout(udp_lookup(qname, qtype.clone(), local), timeout)
+        .await
+        .unwrap_or(Message::new());
+    let remote_response = try_timeout(socks5_lookup(qname, qtype.clone(), socks5, remote), timeout)
+        .await
+        .unwrap_or(Message::new());
```
Mygod: We should send requests first, then await `local_response` and if we know we are going to bypass, we can cancel `remote_response`.

Mygod: Of course, this is lower priority...

src/relay/dnsrelay/mod.rs
Mygod: Also lower priority...

src/context.rs
```diff
@@ -268,12 +282,38 @@ impl Context {
...
+                                if a.check_target_bypassed(self, &reverse_addr).await {
+                                    return true;
+                                }
+                            }
+                        }
+                        _ => (),
+                    }
+                }
+                a.check_target_bypassed(self, target).await
+            }
```
Mygod: `check_target_bypassed` should return `Option<bool>`, and return `None` if nothing matches for hostname.

src/relay/dnsrelay/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+
+    let addr = Address::DomainNameAddress(qname.to_string(), 0);
+    let qname_bypassed = context.check_target_bypassed(&addr).await;
```
Mygod: `qname_bypassed` should account for `Option<bool>` as mentioned elsewhere...

Mygod:
Re @zonyitoo: in general it is helpful to run a socksifier VPN if we run DNS resolver in the same process as sslocal. If you do not plan to support other platforms for now, I don't see an issue with protecting everything with `cfg`.

Mygod:
Also please keep thing in a pull request so it is easier for me to review them and learn rust. :)

zonyitoo:
> If you do not plan to support other platforms for now

Yes, I have no plan for that.

I have made some modification for cleanup in master branch.

src/bin/local.rs
zonyitoo: Yes, this is not used anywhere in code. What is this for? @madeye

Mygod:
That is what I am saying. Please keep the new features in a pull request, this could help me a lot but I guess it is already too late...

We are supposed to use `local_dns_path` instead of `local_dns_addr` on Android so I guess it is still WIP.

zonyitoo:
> I don't see an issue with protecting everything with `cfg`

It is ok to protect it with `cfg`. What I am trying to say is that: we don't need a sepecial feature `dns-relay` to protect that, `target_os = "android"` is already good enough.

madeye:
Okay, I will make this a Android only feature.

src/config.rs
madeye: yeah, fixed.

src/bin/local.rs
madeye: it's for Mygod's UNIX domain socket based local resolver.

I'll apply `cfg!(target_os = "android")` here.

BTW, I think local socket with UDP (in kernel, it's also implemented like UNIX socket) is already good enough, maybe UNIX socket is not necessary?

src/relay/dnsrelay/mod.rs
madeye: Fixed.

PR #207, opened 2020-03-14
Refine China DNS function.

madeye:


zonyitoo:
I have resolved conficts with the master, please double checks @madeye

madeye:
@zonyitoo It's possible that remote DNS address is a domain name. So, please follow my change here, not revert to SocketAddr.

zonyitoo:
Nope. I haven't revert that. I changed to `Address` instead of `String`.

src/config.rs
```diff
@@ -911,7 +911,7 @@ pub struct Config {
     pub local_dns_addr: Option<SocketAddr>,
     /// Remote DNS's address
     #[cfg(target_os = "android")]
-    pub remote_dns_addr: Option<String>,
+    pub remote_dns_addr: Option<Address>,
```
madeye: Should be a `String` here, or `socks5::Address`.

The remote DNS address would be something like: `dns.google.com:53`

zonyitoo: Yes, it is an `socks5::Address`.

src/plugin/ss_plugin.rs
```diff
@@ -22,6 +22,13 @@ pub fn plugin_cmd(plugin: &PluginConfig, remote: &ServerAddr, local: &SocketAddr
 
     if let Some(ref opt) = plugin.plugin_opt {
         cmd.env("SS_PLUGIN_OPTIONS", opt);
+        #[cfg(target_os = "android")]
+        {
+            // Add VPN flags to the commandline as well
+            if opt.contains(";V") {
+                cmd.arg("-V");
+            }
+        }
```
zonyitoo: In shadowsocks-libev:

```c
#ifdef __ANDROID__
    extern int vpn;
    if (vpn)
        cork_exec_add_param(exec, "-V");
#endif
```

I think plugin should read this flag from `SS_PLUGIN_OPTIONS` instead of `-V`, since many programs uses `-V` for printing version.

madeye:
Okay, I cleaned up the commits. Everything should be fine now.

src/plugin/ss_plugin.rs
madeye: It's a workaround for v2ray-plugin, it currently only accepts command line args. Maybe we can apply it on v2ray-plugin only to avoid other potential issues.

zonyitoo:
Allowing setting VPN mode with opts: https://github.com/shadowsocks/v2ray-plugin/pull/193
@madeye

PR #211, opened 2020-03-15
Use local_dns_path

Mygod:
Ran out of time today... I will post this WIP here just in case any of you want to continue refactoring the code... :)


src/relay/dnsrelay/upstream.rs
```diff
@@ -0,0 +1,59 @@
+use std::io::Result;
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, UdpSocket};
+use trust_dns_proto::op::Message;
+
+pub trait Upstream {
+    async fn lookup(&self, request: &Message) -> Result<Message>;
```
zonyitoo: Why your `trait` can have `async fn`? [Shock]

src/relay/dnsrelay/upstream.rs
```diff
@@ -0,0 +1,59 @@
...
+
+impl Upstream for TcpUpstream {
+    async fn lookup(&self, request: &Message) -> Result<Message> {
+        unimplemented!()
+    }
+}
+
+#[cfg(any(target_os = "linux", target_os = "android"))]
+struct UnixSocketUpstream {
+    path: String,
```
zonyitoo: You may want to use `std::path::PathBuf`.

src/relay/dnsrelay/upstream.rs
Mygod: I don't know a lot about rust and this is a WIP. At least we can do this in Kotlin. 🤣 

So I guess it should be `fn lookup(&self, request: &Message) -> Future<Result<Message>>`?

zonyitoo: Use a helper macro: `#[async_trait]`

madeye:
`local_dns_path` works now. @Mygod may want to do some refactoring.

Mygod:
Yes... Please do wait for me to catch up...

madeye:
After days of local testing, I think shadowsocks-rust works properly on Android now. No function issue or memory leak observed.

Let's review this change again and hope it get merged soon.

zonyitoo:
LGTM.

Mygod:
I think we can merge this for now and worry about features later. :)

Issue #213, opened 2020-03-15
DnsRelay improvements


zonyitoo:
- [ ] Allow DNS relay to work as a standalone server
- [ ] Uses the same UDP socket to send reponse packet instead of creating one for every query [related code](https://github.com/shadowsocks/shadowsocks-rust/blob/master/src/relay/dnsrelay/mod.rs#L241)
- [ ] Send DNS query directly to remote `ssserver` instead of sending to another socks5 `sslocal` process

zonyitoo:
Asking for comments @madeye @Mygod .

Reverse IP ACL lookup can be enabled by a separated feature, as suggested in #210 .

zonyitoo:
Question: Why DNS queries are proxied to `remote_dns_addr` with TCP protocol but UDP to `local_dns_addr`? @madeye

Mygod:
It should be configurable :)

madeye:
Many shadowsocks servers don't enable UDP relay, so we forward the DNS query via TCP to the remote server.

zonyitoo:
> Many shadowsocks servers don't enable UDP relay, so we forward the DNS query via TCP to the remote server.

So it should be configurable then.

zonyitoo:
Ok, next step, 

- [ ] making DNS queries sent by TCP/UDP configurable

zonyitoo:
Should we reuse [trust-dns-client](https://crates.io/crates/trust-dns-client) to implement DNS relay? @madeye

madeye:
If possible, it would be very nice.

With trust-dns-client, it maybe possible to handle hosts file as well.

zonyitoo:
But that will end up to link most of the `trust-dns`, binary size will be large.

madeye:
The binary size is not a concern for me.

@Mygod what do you think?

zonyitoo:
Could we reuse the global DNS resolver? @madeye , the one in `src/relay/dns_resolver/`

Mygod:
What are you asking? If global resolver can handle the complexity of things we need to do then sure?

zonyitoo:
`dns_relay` send queries simultaneously by direct connections and proxy connections. For direct connections, they actually work the same as `trust-dns-resolver` internally but lack of many other things, such as lookup strategies (ipv4only, ipv6only, ipv4thenipv6, ...). So I was thinking about we should reuse this library rather than implementing all by ourselves.

[...]

PR #214, opened 2020-03-16
DNS relay refactor and improvements

zonyitoo:
- Allow working in standalone mode
- Use same UDP socket for sending response packet
- Send TCP queries with ProxyStream directly

ref #213


zonyitoo:
Plus: Adding a `local-flow-stat` feature to enable local statistic report.

madeye:
I added some small fixes for clippy via 7ec11df.

PR #235, opened 2020-05-04
Refine DNS relay

Mygod:
Beware of rookie mistakes lol.


Mygod:
Wait why is `dnsrelay/mod` not barred behind `local-dns-relay` feature? I am confused.

zonyitoo:
It should be added behind `local-dns-relay` feature.

src/relay/mod.rs
```diff
@@ -1,7 +1,8 @@
 //! Relay server in local and server side implementations.
 
-pub mod dns;
+// pub mod dns;
```
Mygod, line 3: This `dns` mod seems unused. Is it legacy module that is no longer used?

Mygod:
I have more changes in mind but we can review this first. Changes in this PR:

* Respect both lists for qname matching;
* Only one DNS probe will be sent if qname was matched -- this improves privacy somewhat;
* Reverse lookup cache remembers local/remote instead of qname;
* Some optimizations.

Again beware of my rookie mistakes lol.

src/relay/mod.rs
zonyitoo: There is a binary `ssdns`, which have been removed long time ago.

Mygod: Thanks! Ready to review now.

zonyitoo:
Please merge master again.

Mygod:
GitHub seems to think there are no conflicts?

zonyitoo:
It said: This branch cannot be rebased due to conflicts

Mygod:
Ah I guess rebasing is different. Let me squash everything to a single commit then.

madeye:
This change breaks GFW list and Custom list feature in shadowsocks-android.

I'm debugging what's wrong here.

src/relay/dnsrelay/mod.rs
```diff
@@ -123,65 +122,70 @@ async fn acl_lookup(
...
-            RData::A(ref ip) => {
-                let addr = Address::SocketAddress(SocketAddr::new(IpAddr::from(*ip), 0));
-                context.check_target_bypassed(&addr).await
-            }
-            RData::AAAA(ref ip) => {
-                let addr = Address::SocketAddress(SocketAddr::new(IpAddr::from(*ip), 0));
-                context.check_target_bypassed(&addr).await
+    let qname_in_proxy_list = context.check_qname_in_proxy_list(&addr);
+
+    let remote_response_fut = async {
```
Mygod, line 138: I forgot to invoke `tokio::spawn` here. Is there any way to do this without inappropriate lifetimes?

Mygod: All the scoped libraries I found seems to swallow return values of the future, which is undesirable...

madeye: I think previous join approach is better?

Mygod: We do not want to send out both probes all the time though. If the hostname is matched, certainly we only need to send one probe. This improves privacy slightly (use case: you want to hide even DNS traffic to some sensitive website).

Mygod: This PR looks promising, alas it is a WIP: https://github.com/tokio-rs/tokio/pull/2153/files#diff-d0c8061133c518b906f05037bd563ed7R127

madeye: Hmm, but if we spawn here, actually we send the remote request ahead as well?

Mygod: @madeye Look at the next two lines below.

madeye: What about this?

PR #238, opened 2020-05-07
Fix bugs in #235

madeye:
1. Fix the return value of `acl::check_ip_in_proxy_list`.
2. Fix the condition order in `context::check_ip_in_proxy_list`
3. Keep the value whether the qname is resolved by remote or local in `add_to_reverse_lookup_cache`

Tested locally, all routes in Android client should work now.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -291,9 +291,6 @@ impl Context {
     /// Add a record to the reverse lookup cache
     #[cfg(feature = "local-dns-relay")]
     pub fn add_to_reverse_lookup_cache(&self, addr: &IpAddr, forward: bool) {
-        if self.check_ip_in_proxy_list(addr) == forward {
-            return;
-        }
         let mut reverse_lookup_cache = self.reverse_lookup_cache.lock();
         reverse_lookup_cache.insert(addr.clone(), forward);
     }
@@ -314,13 +311,21 @@ impl Context {
             // Proxy everything by default
             None => true,
             Some(ref a) => {
-                // do the reverse lookup in our local cache
-                let mut reverse_lookup_cache = self.reverse_lookup_cache.lock();
-                // if a qname is found
-                if let Some(forward) = reverse_lookup_cache.get(ip) {
-                    !*forward
+                // first check the IP list
+                if a.check_ip_in_proxy_list(ip) {
+                    true
                 } else {
-                    a.check_ip_in_proxy_list(ip)
+                    // do the reverse lookup in our local cache
+                    let mut reverse_lookup_cache = self.reverse_lookup_cache.lock();
+                    // if a qname is found
+                    if let Some(forward) = reverse_lookup_cache.get(ip) {
+                        // if qname is resolved by remote,
+                        // we should proxy it as well
+                        *forward
+                    } else {
+                        // by default not proxied
+                        false
+                    }
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


src/context.rs
```diff
@@ -314,13 +311,21 @@ impl Context {
...
             None => true,
             Some(ref a) => {
-                // do the reverse lookup in our local cache
-                let mut reverse_lookup_cache = self.reverse_lookup_cache.lock();
-                // if a qname is found
-                if let Some(forward) = reverse_lookup_cache.get(ip) {
-                    !*forward
+                // first check the IP list
+                if a.check_ip_in_proxy_list(ip) {
+                    true
```
Mygod, line 316: This would lead to inconsistent behavior as `check_ip_in_proxy_list` does not distinguish whether some rule is matched or default rule is used.

src/context.rs
```diff
@@ -291,9 +291,6 @@ impl Context {
     /// Add a record to the reverse lookup cache
     #[cfg(feature = "local-dns-relay")]
     pub fn add_to_reverse_lookup_cache(&self, addr: &IpAddr, forward: bool) {
-        if self.check_ip_in_proxy_list(addr) == forward {
-            return;
-        }
```
Mygod, line 296: This optimization should probably be fine?

madeye: It's possible that the domain name is marked proxied, but IP is not in the proxy list. Typically in GFW List mode.

src/context.rs
madeye: For IP rules, I think this ensures `proxy_list` in `bypass_all` and `bypass_list` in `proxy_all` have the highest priority. Otherwise we would see default domain name rules overwrite IP rules in those modes.

We can think about this, as there maybe better solutions.

src/context.rs
Mygod: Hmm maybe we should instead check whether the outcome is consistent with the cache and add/remove from cache accordingly. (I want to minimize memory footprint/maximize LruCache)

src/context.rs
Mygod: I think cache should take precedence as cache directly corresponding to a hostname match.

madeye: IMO, any hostname resolved by DNS relay should be cached, otherwise we cannot apply hostname rules to all of them.

src/context.rs
madeye: IMO, any hostname resolved by DNS relay should be cached, otherwise we cannot apply hostname rules to all of them.

src/context.rs
madeye: Yeah, it maybe doable.

madeye:
Oops, it seems I pushed to the wrong repo...

src/context.rs
madeye: https://github.com/shadowsocks/shadowsocks-rust/commit/ee1fd01698ca807e2aa39431e0a5b7b37058dac8

madeye:
@zonyitoo you may want to do a force push to redo my merge...

src/context.rs
Mygod: That I agree. These three lines are my attempt at an optimization (it should not change behavior at all).

Mygod:
https://github.com/shadowsocks/shadowsocks-rust/compare/b6c3fb0babd621a2fc6749931f3a140e5833a693...master

So I guess the real bug is flipping proxy/bypass list it seems.

src/context.rs
madeye: Okay, I think I can update the cache in `check_ip_in_proxy_list` to solve this.

madeye:
@Mygod Okay, I think you're right.

zonyitoo:
WHAT happened...

Mygod:
@zonyitoo You didn't see anything.

PR #239, opened 2020-05-07
Fix bugs in #235

madeye:
Details in https://github.com/shadowsocks/shadowsocks-rust/pull/238


madeye:
> Although (completely unrelated to changes currently) maybe we should remove cache entries when adding cache

Let's make this a future improvement, although this use case doesn't make sense to me...

Mygod:
It's an edge case rather than a use case to be honest XD (it offers better consistency rather than utility).

zonyitoo:
Actually I don't know why Android needs so many complicated modification :P

PR #245, opened 2020-05-10
compatible with proxy_list and bypass_list, in the one ACL file. fix compilation errors

dev4u:
fix compilation errors, caused by https://github.com/shadowsocks/shadowsocks-rust/commit/7a698e36e7a084bf9490f51fff0f64a92e9b4b3d

### DIFF
--- a/src/acl/mod.rs
+++ b/src/acl/mod.rs
@@ -281,7 +281,7 @@ impl AccessControl {
     /// If so, it should be resolved from remote (for Android's DNS relay)
     pub fn check_host_in_proxy_list(&self, host: &str) -> Option<bool> {
         match self.mode {
-            //"[accept_all]" | "[proxy_all]"
+            // "[accept_all]" | "[proxy_all]"
             Mode::BlackList => {
                 //
                 // [proxy_all]
@@ -302,7 +302,7 @@ impl AccessControl {
                     return Some(false);
                 }
             },
-            //"[reject_all]" | "[bypass_all]"
+            // "[reject_all]" | "[bypass_all]"
             Mode::WhiteList => {
                 //
                 // [bypass_all]
@@ -343,20 +343,8 @@ impl AccessControl {
 
     pub fn check_ip_in_proxy_list(&self, ip: &IpAddr) -> bool {
         match self.mode {
-            //"[accept_all]" | "[proxy_all]" see also #check_host_in_proxy_list()
-            Mode::BlackList => {
-                if self.white_list.check_ip_matched(ip) {
-                    return true;
-                }
-                return !self.black_list.check_ip_matched(ip)
-            },
-            //"[reject_all]" | "[bypass_all]" see also #check_host_in_proxy_list()
-            Mode::WhiteList => {
-                if self.black_list.check_ip_matched(ip) {
-                    return false;
-                }
-                return self.white_list.check_ip_matched(ip)
-            },
+            Mode::BlackList => !self.black_list.check_ip_matched(ip),
+            Mode::WhiteList => self.white_list.check_ip_matched(ip),
         }
     }
 
--- a/src/acl/mod.rs
+++ b/src/acl/mod.rs
@@ -9,7 +9,7 @@ use std::{
     net::{IpAddr, SocketAddr},
     path::Path,
 };
-
+use log::info;
 use ipnet::{IpNet, Ipv4Net, Ipv6Net};
 use iprange::IpRange;
 use regex::{RegexSet, RegexSetBuilder};
@@ -280,13 +280,50 @@ impl AccessControl {
     /// Check if domain name is in proxy_list.
     /// If so, it should be resolved from remote (for Android's DNS relay)
     pub fn check_host_in_proxy_list(&self, host: &str) -> Option<bool> {
-        // Addresses in proxy_list will be proxied
-        if self.white_list.check_host_matched(host) {
-            return Some(true);
-        }
-        // Addresses in bypass_list will be bypassed
-        if self.black_list.check_host_matched(host) {
-            return Some(false);
+        info!("host checking: {}", host);
+        match self.mode {
+            Mode::BlackList => { //"[accept_all]" | "[proxy_all]"
+                //
+                // [proxy_all]
+                // [proxy_list]
+                // (?:^|\.)cname\.example\.com$
+                // [bypass_list]
+                // (?:^|\.).example\.com$
+                //
+                // access: cname.example.com  => proxy
+                // access: (**).example.com   => direct
+                //
+                // Addresses in proxy_list will be proxied
+                if self.white_list.check_host_matched(host) {
+                    info!("host: {}, in BL mode, exists in white_list, return: true(in proxy)", host);
+                    return Some(true);
+                }
+                // Addresses in bypass_list will be bypassed
+                if self.black_list.check_host_matched(host) {
+                    info!("host: {}, in BL mode, exists in back_list, return: false(no proxy)", host);
+                    return Some(false);
+                }
+            },
+            Mode::WhiteList => { //"[reject_all]" | "[bypass_all]"
+                //
+                // [bypass_all]
+                // [bypass_list]
+                // (?:^|\.)cname\.example\.com$
+                // [proxy_list]
+                // (?:^|\.).example\.com$
+                //
+                // access: cname.example.com  => direct
+                // access: (**).example.com   => proxy
+                //
+                if self.black_list.check_host_matched(host) {
+                    info!("host: {}, in WL mode, exists in black_list, return: false(no proxy)", host);
+                    return Some(false);
+                }
+                if self.white_list.check_host_matched(host) {
+                    info!("host: {}, in WL mode, exists in white_list, return: true(in proxy)", host);
+                    return Some(true);
+                }
+            },
         }
         None
     }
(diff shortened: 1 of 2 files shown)


zonyitoo:
Only `context.rs` has compliation error.

But modifications in `mod.rs` doesn't seem to be related.

src/acl/mod.rs
```diff
@@ -280,13 +280,50 @@ impl AccessControl {
...
-        // Addresses in proxy_list will be proxied
-        if self.white_list.check_host_matched(host) {
-            return Some(true);
-        }
-        // Addresses in bypass_list will be bypassed
-        if self.black_list.check_host_matched(host) {
-            return Some(false);
+        info!("host checking: {}", host);
+        match self.mode {
+            Mode::BlackList => { //"[accept_all]" | "[proxy_all]"
```
zonyitoo, line 285: Code style:

1. this line of comment should be in a separated line
2. an empty space is required after `//`

```rust
// "[accept_all]" | "[proxy_all]"
Mode::BlackList => {
```

src/acl/mod.rs
```diff
@@ -280,13 +280,50 @@ impl AccessControl {
...
+                    info!("host: {}, in BL mode, exists in white_list, return: true(in proxy)", host);
+                    return Some(true);
+                }
+                // Addresses in bypass_list will be bypassed
+                if self.black_list.check_host_matched(host) {
+                    info!("host: {}, in BL mode, exists in back_list, return: false(no proxy)", host);
+                    return Some(false);
+                }
+            },
+            Mode::WhiteList => { //"[reject_all]" | "[bypass_all]"
```
zonyitoo, line 307: Same as above.

src/acl/mod.rs
```diff
@@ -309,8 +346,18 @@ impl AccessControl {
 
     pub fn check_ip_in_proxy_list(&self, ip: &IpAddr) -> bool {
         match self.mode {
-            Mode::BlackList => !self.black_list.check_ip_matched(ip),
-            Mode::WhiteList => self.white_list.check_ip_matched(ip),
+            Mode::BlackList => { //"[accept_all]" | "[proxy_all]" see also #check_host_in_proxy_list()
```
zonyitoo, line 349: Same as above.

src/acl/mod.rs
```diff
@@ -309,8 +346,18 @@ impl AccessControl {
...
+                if self.white_list.check_ip_matched(ip) {
+                    return true;
+                }
+                return !self.black_list.check_ip_matched(ip)
+            },
+            Mode::WhiteList => { //"[reject_all]" | "[bypass_all]" see also #check_host_in_proxy_list()
```
zonyitoo, line 355: Same as above.

zonyitoo:
On the other hand, @madeye please confirm that modifications of ACL matching rules remain consistent with `shadowsocks-libev`.

dev4u:
it's my fault. i'm use PR for the first time……
the mod.rs is another topic.

zonyitoo:
It's ok. And I suggests the first line of commit message should contain a brief of what have you modified.

It's also ok for you to open another PR for ACL rules. Otherwise, you can just modify your PR's title.

dev4u:
> It's ok. And I suggest the first line of commit message should contain a brief of what have you modified.
> 
> It's also ok for you to open another PR for ACL rules. Otherwise, you can just modify your PR's title.

thank you for your relay.

src/acl/mod.rs
zonyitoo: an empty space after `//` is not added yet.

src/acl/mod.rs
```diff
@@ -280,9 +280,9 @@ impl AccessControl {
     /// Check if domain name is in proxy_list.
     /// If so, it should be resolved from remote (for Android's DNS relay)
     pub fn check_host_in_proxy_list(&self, host: &str) -> Option<bool> {
-        info!("host checking: {}", host);
         match self.mode {
-            Mode::BlackList => { //"[accept_all]" | "[proxy_all]"
+            //"[accept_all]" | "[proxy_all]"
```
zonyitoo, line 284: Add an empty space after `//`.

src/acl/mod.rs
```diff
@@ -309,8 +343,20 @@ impl AccessControl {
...
+                }
+                return !self.black_list.check_ip_matched(ip)
+            },
+            //"[reject_all]" | "[bypass_all]" see also #check_host_in_proxy_list()
+            Mode::WhiteList => {
+                if self.black_list.check_ip_matched(ip) {
+                    return false;
+                }
+                return self.white_list.check_ip_matched(ip)
+            },
```
Mygod, line 359: Only one IP list will be in effect. This is intended behavior.

madeye:
LGTM.

zonyitoo:
`check_host_in_proxy_list` is not reverted yet.

Mygod:
Technically the test cases supplied there should cause undefined behavior (you could use regex rules to exclude the exceptions), but I think the proposal makes sense and the overhead is small.

zonyitoo:
If only one IP list will be effected, why should we allow two host lists?

Mygod:
Because the outcome for determining bypassing IP is binary -- either proxy or bypass.

For hostnames however, since DNS is just an auxiliary to the IP stack, we determine the outcome based on both hostname and its resolved IPs, so the outcome is ternary -- proxy, bypass, or inconclusive, where we decide the final outcome basing on resolved IPs.

dev4u:
i.g., in China, we often use the whitelist mode, by specifying which sites need to be accessed through a proxy.  And there will be a problem, such as the gstatic.com site, which is prohibited from access on the mainland.  However, there are individual subdomains that allow direct access.  At this time I can configure my ACL file like this:
[bypass_all]
[bypass_list]
connectivity.gstatic.com
[proxy_list]
gstatic.com

so In this way, I guarantee both the access to gstatic.com and the quality of the service.

Mygod:
@dev4u You should use negative lookbehind to do this, e.g. `(?:^|\.)(?<!^connectivity\.|\.connectivity\.)gstatic.com`.

dev4u:
Of course you can do this, but it is not so easy to maintain such an acl file through the community.  Because subdomains are indefinite, some need proxy, and some need bypass, the person who amends may not know how to modify the rules he wants.

dev4u:
@Mygod Of course you can do this, but it is not so easy to maintain such an acl file through the community. Because subdomains are indefinite, some need proxy, and some need bypass, the person who amends may not know how to modify the rules he wants.

zonyitoo:
> Because the outcome for determining bypassing IP is binary -- either proxy or bypass.

Yes its true. But I think the main concern here is about priority.

If an IP / Hostname matches rules in both `[proxy_list]` and `[bypass_list]`, we could decide that which list we checked first by `[bypass_all]` or `[proxy_all]` (in other word, `mode`).

So the result is always conclusive. If `[proxy_list]` overlaps with `[bypass_list]`, priority is decided by `[proxy_list]` and `[bypass_list]`.

Hostnames could be inconclusive if it is not defined in lists, but ... it is expected for all users if they rely on this behavior.

Mygod:
@dev4u The solution you proposed is hacky and only handles double exceptions (the exception to exceptions to the general rules). Also I am not convinced that `connectivitycheck.gstatic.com` is a useful case to support. Either way the exceptions should be very rare.

@zonyitoo The primary use case I think @madeye want ACL to support is to (a) bypass LAN and China (or more general, local area network) and (b) handle DNS pollution. For (a) the easiest way is to resolve hostname and bypass them if needed; for (b) we want the hostname rules to take priority.

madeye:
Right, hostname rules always have the highest priority.

dev4u:
Sorry, the example I gave may not be a good case to support why that is necessary.
The thing I want to explain is that if there are exceptions in the acl file, the experience of using the network will be improved.

On the mainland, as @Mygod said, using bypass LAN and China mode will route a considerable portion of browsing requests to vps as a proxy. This is because to avoid dns pollution, as shadowsocks's designs, dns requests to be forwarded to vps for resolution (this is very good), which directly leads to a high probability that the resolved ip is not in Chinese IP (because of cdn). 

For example, if the download request of the Xiaomi application market is resolved overseas, it will be resolved to Singapore's IP. (p.s., this is because you want to download the google play application, you need to add the download manager to the proxy list.)
In a complicated network usage environment in China, a simple routing rule makes it difficult to have a good user experience. Just as gfw intercepted us in multiple dimensions, we have to change our strategy accordingly so as not to be beaten.

zonyitoo:
Alright. Then ACL matching logic in shadowsocks-libev should be modified to allow matching hostnames with priority. It's about consistency of all implementations.

zonyitoo:
> Right, hostname rules always have the highest priority.

The *priority* I was talking about is: when a IP or HOSTNAME matches rules in both `[proxy_list]` and `[bypass_list]`, what should we do? Bypass or proxy?

```ini
[proxy_list]
127.0.0.1/8
(?:^|\.)\.example\.com

[bypass_list]
127.0.0.1/16
(?:^|\.)\.example\.com
```

Because IP rules are in CIDR form, and HOSTNAME rules are regular expressions, there must be overlapped rules in `proxy_list` and `bypass_list`.

I personally prefer @dev4u 's original solution, both `check_ip_in_proxy_list` and `check_host_in_proxy_list` check both lists. For `BlackList` mode, check `white_list` first, and for `WhiteList` mode, check `black_list` first.

dev4u:
@zonyitoo Okay, after dawn, I will make another PR to fixing the compilation error.

Mygod:
@zonyitoo As I proposed, this in principle should be undefined behavior and it is user's responsibility to prevent this from happening. (trivial to do for IP blocks, for hosts do something like https://github.com/shadowsocks/shadowsocks-rust/pull/245#issuecomment-626950705)

If you really want to go overboard with @dev4u's proposal, why stop at double exception when there might be triple exceptions? For example, if I want to proxy `*.a`, but not `*.b.a`, but I do want to proxy `*.c.b.a`?

Mygod:
@dev4u Did you read the resolving logic I wrote? Two probes (one local, one remote) will be sent for every unmatched hostname which takes care of the issue you mentioned.

zonyitoo:
Yes, you are right.

dev4u:
> @dev4u Did you read the resolving logic I wrote? Two probes (one local, one remote) will be sent for every unmatched hostname which takes care of the issue you mentioned. In particular, the outcome depends on local response, not remote response.

In fact, I have not read this piece of processing logic. If my comments are inappropriate, I am sorry.
As I mentioned above, I just use the appearance to explore the logic processing in shadowsocks.
The ss I used is gfw mode, dns is a mainland private dns server (dot), and the upstream dns address is 119.29. **

PR #246, opened 2020-05-11
Let DNS relay run on TCP as well

Mygod:
I tried integrating `trust-dns-server`, it turns out that using it will block the entire thread (it takes a `&Runtime` as input). To be fair, I suspect there is anything we can really use from trust-dns besides `trust-dns-proto`.

P.S. This TCP server closes connection for every query. This should not be an issue as the relay should run locally.

### DIFF
--- a/src/relay/dnsrelay/mod.rs
+++ b/src/relay/dnsrelay/mod.rs
@@ -28,7 +28,10 @@ pub mod upstream;
 
 fn should_forward_by_ptr_name(acl: &AccessControl, name: &Name) -> bool {
     let mut iter = name.iter().rev();
-    let mut next = || std::str::from_utf8(iter.next().unwrap_or(&[48])).unwrap_or("*");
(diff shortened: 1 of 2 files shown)


zonyitoo:
Ah, so you are going to make dns-relay to be a DNS server?

Mygod:
It has always been a DNS server, no? A server that only relays other's messages but a server nonetheless.

Mygod:
It seems like build was failing due to missing fix in #245. Any other comments? @zonyitoo

dev4u:
> It seems like build was failing due to missing fix in #245. Any other comments? @zonyitoo

@zonyitoo pls. review #248 . if there is no problem, pls merge to master.

zonyitoo:
Merged. Please rebase with master.

Mygod:
Looks like all checks have passed.

Issue #252, opened 2020-05-14
File descriptor leak on Linux


cfal:
I'm running shadowsocks 1.8.11 and have also seen this in older versions.

Every week or so, I see the following error:

`thread 'main' panicked at 'create udp association: Os { code: 24, kind: Other, message: "No file descriptors available" }', /volume/src/relay/udprelay/server.rs:371:21`

I previously ran shadowsocks-libev on the same server and never encountered this problem, so believe the issue is with this implementation.

Command line:
```
ssserver --no-delay -c config.json
```

Config:
```
{
    "server": "...",
    "server_port": ...,
    "password": ".."
    "timeout": 300,
    "method": "chacha20-ietf-poly1305",
    "mode":"tcp_and_udp"
}
```

zonyitoo:
`shadowsocks-rust` uses a cache with expiry duration. By default all UDP associations are kept for **5 minutes**, which should be the cause of your issue.

https://github.com/shadowsocks/shadowsocks-rust/blob/master/src/relay/udprelay/server.rs#L285

`shadowsocks-libev` uses a LRU with limited size. `MAX_UDP_CONN_NUM` is `512` on the server side.

https://github.com/shadowsocks/shadowsocks-libev/blob/master/src/udprelay.c#L1414

So that's why you haven't encountered this problem when you are using the `libev` version.

I personally prefer the cache with expiry duration, because you can always increase your system's file descriptor limit to hold as many UDP associations as possible.

zonyitoo:
You may try to add a timeout in your configuration, for example:

```json
{
    "udp_timeout": 10
}
```

This will tells `ssserver` to keep idling UDP associations for **10 seconds**.

cfal:
Thanks for the quick response. I've changed that line to initialize LruCache using with_capacity(512) to see if that solves the problem. Since it seems lru_time_cache::LruCache already supports both capacity and expiry, would a pull request to make this configurable be accepted?

zonyitoo:
Yes, sure.

zonyitoo:
You may need to change both server and local implementations.

PR #253, opened 2020-05-14
Add option for max number of UDP associations

cfal:
https://github.com/shadowsocks/shadowsocks-rust/issues/252

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -102,6 +102,8 @@ struct SSConfig {
     #[serde(skip_serializing_if = "Option::is_none")]
     udp_timeout: Option<u64>,
     #[serde(skip_serializing_if = "Option::is_none")]
+    udp_max_conns: Option<u64>,
+    #[serde(skip_serializing_if = "Option::is_none")]
     servers: Option<Vec<SSServerExtConfig>>,
     #[serde(skip_serializing_if = "Option::is_none")]
     dns: Option<String>,
@@ -911,6 +913,8 @@ pub struct Config {
     pub config_type: ConfigType,
     /// Timeout for UDP Associations, default is 5 minutes
     pub udp_timeout: Option<Duration>,
+    /// Maximum number of UDP Associations, default is unconfigured
+    pub udp_max_conns: Option<u64>,
     /// `RLIMIT_NOFILE` option for *nix systems
     pub nofile: Option<u64>,
     /// Timeout for TCP connections, could be replaced by server*.timeout
@@ -1019,6 +1023,7 @@ impl Config {
             manager_method: None,
             config_type,
             udp_timeout: None,
+            udp_max_conns: None,
             nofile: None,
             timeout: None,
             acl: None,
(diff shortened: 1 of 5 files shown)


src/config.rs
```diff
@@ -102,6 +102,8 @@ struct SSConfig {
     #[serde(skip_serializing_if = "Option::is_none")]
     udp_timeout: Option<u64>,
     #[serde(skip_serializing_if = "Option::is_none")]
+    udp_max_conns: Option<u64>,
```
zonyitoo, line 105: I would prefer `udp_max_associations`, as your comment.

src/relay/udprelay/redir_local.rs
```diff
@@ -94,10 +94,16 @@ pub async fn run(context: SharedContext) -> io::Result<()> {
...
     info!("shadowsocks UDP redirect listening on {}", local_addr);
 
-    // NOTE: Associations are only eliminated by expire time
+    // NOTE: Associations are only eliminated by expire time by default
     // So it may exhaust all available file descriptors
     let timeout = context.config().udp_timeout.unwrap_or(DEFAULT_TIMEOUT);
-    let assoc_map: SharedAssocMap = Arc::new(Mutex::new(LruCache::with_expiry_duration(timeout)));
+    let max_conns = context.config().udp_max_conns.unwrap_or(0) as usize;
+
+    let assoc_map: SharedAssocMap = Arc::new(Mutex::new(if max_conns > 0 {
```
zonyitoo, line 102: This is bad, why not

```rust
let max_assoc_opt = context.config().udp_max_associations;
if let Some(ma) = max_assoc_opt {
    LruCache::with_expiry_duration_and_capacity(timeout, ma)
} else {
    LruCache::with_expiry_duration(timeout)
}
```

zonyitoo: Same in the others.

zonyitoo:
```rust
let assoc_map = if let ...
let assoc_map = Arc::new(Mutex::new(assoc_map));
```

cfal: If done as per your first comment, then it's possible for a user to configure udp_max_associations = 0. I'm not certain as to the behavior of the cache were that to happen. Adding an additional bounds check for that case inside the if let clause would make the code even longer. Which is why I chose to unwrap_or(0) first and then do the bounds checking. What do you think?

zonyitoo: `udp_max_associations = 0` should be checked as an error and report it to user.

If user have explictly specify a value for this configure option, you should use it instead of making one of the value to be the same as not configured.

zonyitoo:
@cfal Could you push your fixes?

zonyitoo:
I am going to merge it and change the name of field.

cfal:
Thanks for finishing this. And sorry for the delayed response.

PR #255, opened 2020-05-17
Refine support for MX and NS records in DNS relay

Mygod:
Added additional data examination for MX and NS records as per RFC. Probably will not be useful ever. Untested.


zonyitoo:
That's strange. Travis-CI is broken.

PR #258, opened 2020-05-20
Warn when using tokio DNS resolver

### LINKED ISSUE #208: tokio_dns_resolver is actually blocking
https://github.com/tokio-rs/tokio/blob/826fc21abfd04779cde92e4cc33de2adb42cf327/tokio/src/net/addr.rs#L147-L148

I rest my case.

Mygod:
Fixes #208. Untested.


zonyitoo:
😅Why...

Mygod:
Because blocking = bad?

PR #265, opened 2020-05-23
Connect UDP sockets

### LINKED ISSUE #263: Domain name is re-resolved for every UDP packet
This is *very* wasteful. Consider using `UdpSocket::connect` instead.

See also this for unexpected benefits: https://stackoverflow.com/a/51296247/2245107

Mygod:
Fixes #263. Untested.

### DIFF
--- a/src/relay/dnsrelay/upstream.rs
+++ b/src/relay/dnsrelay/upstream.rs
@@ -137,10 +137,11 @@ impl Upstream for UdpUpstream {
             SocketAddr::V4(..) => IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)),
             SocketAddr::V6(..) => IpAddr::V6(Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 0)),
         }, 0)).await?;
-        socket.send_to(&generate_query_message(query).to_vec()?, self.server).await?;
+        socket.connect(self.server).await?;
+        socket.send(&generate_query_message(query).to_vec()?).await?;
         let mut response = vec![0; 512];
-        socket.recv_from(&mut response).await?;
-        Ok(Message::from_vec(&response)?)
+        let len = socket.recv(&mut response).await?;
+        Ok(Message::from_vec(&response[..len])?)
     }
 }
 
--- a/src/relay/loadbalancing/server.rs
+++ b/src/relay/loadbalancing/server.rs
@@ -474,7 +474,7 @@ impl<S: ServerData + 'static> PingBalancer<S> {
 
         let addr = Address::SocketAddress(SocketAddr::new(IpAddr::V4(Ipv4Addr::new(8, 8, 8, 8)), 53));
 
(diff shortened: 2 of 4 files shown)


zonyitoo:
Looks awesome!

zonyitoo:
Tests failed:

```
[ERROR shadowsocks::relay::udprelay::association] failed to send packet 127.0.0.1:59742 -> 127.0.0.1:9230, bypassed? false, error: Destination address required (os error 89)
```

zonyitoo:
https://github.com/shadowsocks/shadowsocks-rust/pull/265/files#diff-7884f1e88e152d534accbe0bdc05abe6L329-L332

These `send_to` should also be replaced with `send`.

zonyitoo:
I see. You have forgot this one: https://github.com/shadowsocks/shadowsocks-rust/pull/265/files#diff-7884f1e88e152d534accbe0bdc05abe6R154

zonyitoo:
```rust
fn connect_remote(context: &Context, socket: &UdpSocket, svr_cfg: &ServerConfig) {
```

zonyitoo:
@Mygod Come here and fix the test errors.

Mygod:
What is `connect_remote`?

zonyitoo:
> What is `connect_remote`?

I mean you can put [these codes](https://github.com/shadowsocks/shadowsocks-rust/pull/265/files#diff-7884f1e88e152d534accbe0bdc05abe6R80-R88) into a `connect_remote` and call it from `associate_proxied` and `associate_with_acl `

Mygod:
Feel free to do it for me. :)

zonyitoo:
I can't push to your branch.

Mygod:
Well you can... I think I enabled "allow edits from maintainer".

zonyitoo:
Already done locally. :P

Mygod:
Rant: `Arc<Context>` is so hard to work with. We do not need to use this `Arc` if only `tokio` supports non `static spawns. Ugh.

zonyitoo:
> we probably can just use a `&'a Context` pointer everywhere

I was thinking the same from years ago.

Mygod:
How about always creating a `Context` with `'static`/nonmanaged lifetime? e.g. `Box::into_raw(Box::new(Context::new(...)))` (this gives `*mut Context` whose lifetime is unmanaged) This should be acceptable as we never deconstruct `Context` unless on exit/panic (even if it is used as a library).

Another alternative is to support graceful shutdown (support cancellation?) and [`tokio::task::scope`](https://github.com/tokio-rs/tokio/pull/2153).

zonyitoo:
Using `unsafe` will always work.

Mygod:
`Box::into_raw` is a safe method as rust runtime does not guarantee destruction of objects. (in fact the entire snippet there is safe)

On the other hand if we implement cancellation, we can probably wait for plugin process to exit as well since it does not involve panicking the tokio Runtime.

PR #267, opened 2020-06-06
Remove uses of pin_project::project attribute

taiki-e:
pin-project will deprecate the project attribute due to some unfixable
limitations.

Refs: https://github.com/taiki-e/pin-project/issues/225

*Although the deprecation of the project attribute has not been released yet, this PR is submitted in advance to avoid the CI from being broken by deprecated warnings.*

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -119,7 +119,7 @@ lru_time_cache = "0.10"
 hyper = { version = "0.13", optional = true }
 http = { version = "0.2", optional = true }
 tower = { version = "0.3", optional = true }
-pin-project = "0.4"
+pin-project = "0.4.17"
 socket2 = "0.3"
 cfg-if = "0.1"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -119,7 +119,7 @@ lru_time_cache = "0.10"
 hyper = { version = "0.13", optional = true }
 http = { version = "0.2", optional = true }
 tower = { version = "0.3", optional = true }
-pin-project = "0.4"
+pin-project = "0.4.17"
```
zonyitoo, line 122: Why pinning to v0.4.17?

taiki-e: It is because 0.4.17 is the minimal version of the alternative way of project attribute is available. (Also, as it is not "=" requirements, it does not actually pin version of dep)

PR #278, opened 2020-07-16
Fix shadowsocks/shadowsocks-android#2562

madeye:
Fix https://github.com/shadowsocks/shadowsocks-android/issues/2562

### DIFF
--- a/src/plugin/ss_plugin.rs
+++ b/src/plugin/ss_plugin.rs
@@ -21,14 +21,18 @@ pub fn plugin_cmd(plugin: &PluginConfig, remote: &ServerAddr, local: &SocketAddr
         .stdin(Stdio::null());
 
     if let Some(ref opt) = plugin.plugin_opt {
-        cmd.env("SS_PLUGIN_OPTIONS", opt);
-        #[cfg(target_os = "android")]
-        {
-            // Add VPN flags to the commandline as well
-            if opt.contains(";V") {
-                cmd.arg("-V");
-            }
-        }
+		if cfg!(target_os = "android")
+		{
+			let mut tmp = opt.clone();
+			// Add VPN flags to the commandline as well
+			if tmp.contains(";V") {
+				cmd.arg("-V");
+				tmp = tmp.replace(";V", "")
+			}
+			cmd.env("SS_PLUGIN_OPTIONS", tmp);
+		} else {
+			cmd.env("SS_PLUGIN_OPTIONS", opt);
+		}
     }
 
     cmd


Mygod:
> The latest openssl 0.10.30 pakcage requires a minimal SDK target 23

Why is this?

Also there is a better fix imo for this.

zonyitoo:
> The latest openssl 0.10.30 pakcage requires a minimal SDK target 23

That's bad.. :(

Mygod:
https://github.com/shadowsocks/shadowsocks-android/commit/0d75659ce6f2d4a8e125c5d10c53a508b28fcdad

We should try to find a fix for 0.10.30 as well.

madeye:
Maybe we just increase the minSDK to 23...

Mygod:
Where did you find the information regarding minSDK?

RebelliousWhiz:
I don't know what's impacting the code, but to my perspective, people still use SDK 23 devices don't care about GFW.... My grandma has a MI 6, and it's now on Android 10...

I guess it's completely fine to set minSDK to 23.

Mygod:
Reported to upstream: https://github.com/sfackler/rust-openssl/issues/1315

If we do not get a satisfactory fix we can drop support for Android 5.

zonyitoo:
So, is it ok for me to merge this MR? @Mygod

Mygod:
I don't think we need this merged for now.

src/plugin/ss_plugin.rs
```diff
@@ -21,14 +21,18 @@ pub fn plugin_cmd(plugin: &PluginConfig, remote: &ServerAddr, local: &SocketAddr
...
-                cmd.arg("-V");
-            }
-        }
+		if cfg!(target_os = "android")
+		{
+			let mut tmp = opt.clone();
+			// Add VPN flags to the commandline as well
+			if tmp.contains(";V") {
+				cmd.arg("-V");
+				tmp = tmp.replace(";V", "")
```
zonyitoo, line 30: BTW, `;V` parameter is only for v2ray-plugin. This strategy should only be applied to this specific plugin, right?

madeye:
I think the missing part is how to pass VPN option to plugins, which is not defined before.

Maybe we should fix this from plugin's side instead.

src/plugin/ss_plugin.rs
madeye: Not sure how third-party plugins handle these options. But currently we only see v2ray-plugin has issues to parse `;V` options.

Mygod:
So let's settle this once and for all: what should we use? `-V` or `V` in plugin options.

zonyitoo:
Since `SS_PLUGIN_OPTIONS` is defined as "options for plugin", the VPN flag should be passed in options.

zonyitoo:
Or maybe we can make another new environment variable, such as `SS_PLUGIN_CTRL`, which could be used for passing standardized controlling options, like `vpn_mode=1&verbose=3&fast_open=1`

madeye:
Let's put `V` into the `SS_PLUGIN_OPTIONS`. I'll release a new `v2ray-plugin` to ensure the VPU flags is parsed correctly.

madeye:
Fixed via https://github.com/shadowsocks/v2ray-plugin-android/commit/47888430131386edfa7c635e324796ed0357b85b

Mygod:
Should we get rid of this hack then?

https://github.com/shadowsocks/shadowsocks-rust/blob/efd9e7747db5b1f924943cc29e570c877167b476/src/plugin/ss_plugin.rs#L28

madeye:
Yep, let's remove that hack.

PR #279, opened 2020-07-17
Fix the tx/rx update in ProxyStream

madeye:
This reverts commit 205eaa95addbb56fccc3d0d34705a92e79c8086b.

Reported in this bug: https://github.com/shadowsocks/shadowsocks-android/issues/2561

Also fix the tx/rx in ProxyStream

### DIFF
--- a/src/relay/local.rs
+++ b/src/relay/local.rs
@@ -176,8 +176,7 @@ async fn flow_report_task(context: SharedContext) -> io::Result<()> {
         let tx = flow_stat.tcp().tx() + flow_stat.udp().tx();
         let rx = flow_stat.tcp().rx() + flow_stat.udp().rx();
 
-        // first is rx, second is tx.
-        let buf: [u64; 2] = [rx, tx];
+        let buf: [u64; 2] = [tx, rx];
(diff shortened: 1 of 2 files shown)


madeye:
On hold for now.

zonyitoo:
Errr. Is this ready for merge?

madeye:
Let's hold on for now. I'm doing the test...

madeye:
We defined `tx` as `upload` in shadowsocks-android and shadowsocks-libev before.

However, it looks shadowsocks-rust increase the `tx` bytes when `read` bytes from remote (`download`): https://github.com/shadowsocks/shadowsocks-rust/blob/master/src/relay/tcprelay/proxy_stream.rs#L371

madeye:
It looks UDP traffic stat is correct, we call `incr_tx` when send packets to remote: https://github.com/shadowsocks/shadowsocks-rust/blob/master/src/relay/udprelay/association.rs#L349

madeye:
At first, I thought it's an order issue.

However, it looks that we get the tx/rx counted incorrectly in the TCP part of shadowsocks-rust.

.

PR #280, opened 2020-07-23
Remove the VPN option hack

madeye:
Remove the previous hack, according to the discussion here: https://github.com/shadowsocks/shadowsocks-rust/pull/278


Mygod:
Manual approval.

Issue #326, opened 2020-11-26
Split library in separated crates


zonyitoo:
Moved from #199 . Work to do:

- [x] Take ownership of [shadowsocks](https://crates.io/crates/shadowsocks)
- Separate crates into
    - [ ] shadowsocks-core (will be released as `shadowsocks` crate), which contains:
        - [ ] TCP Relay: a Stream like interface for communicating with remote servers (Local + Server)
            - Extra API, connects without `Context`:
              ```rust
              ProxyTcpStream::connect(server_addr, method, password).await
              ```
        - [ ] UDP Relay, an Association Manager, for building a NAT to remote servers (Local + Server)
        - [ ] Plugin
        - [ ] ACL, including libev's ACL list, and other possible implementations, like GeoIP
        - [ ] Config, including basic format parsing, SIP002 URL parsing facilities

    - shadowsocks binaries (will be released as `shadowsocks-rust` crate), which contains:
        - [ ] Logging facilities, currently uses `log4rs`, could be customized
        - [ ] sslocal
            - [ ] socks5, socks4/4a, or universal socks
            - [ ] http/https
            - [ ] tunnel
            - [ ] tun/tap
            - [ ] redir
            - [ ] dns
        - [ ] ssserver
        - [ ] ssmanager
        - [ ] ssurl

    - [x] Crypto Library, [shadowsocks-crypto](https://github.com/shadowsocks/shadowsocks-crypto)

gfreezy:
这个是在哪个分支上？ shadowsocks-crypto 是已经拆出来了吗？

zonyitoo:
> 这个是在哪个分支上？ shadowsocks-crypto 是已经拆出来了吗？

Yes. [shadowsocks-crypto](https://crates.io/crates/shadowsocks-crypto).

gfreezy:
感觉下面每个独立成一个 crate，方便复用

协议相关：
- [ ] socks5
- [ ] socks4/4a
- [ ] http/https
- [ ] shadowsocks

接入方式：
- [ ] tunnel
- [ ] redir

zonyitoo:
So if I want to make a release, I will have to release at least 7 crates.

http/https are implemented with hyper, anyone who wants to write a HTTP proxy could use hyper from scratch.

`shadowsocks-local` is good enough, because shadowsocks' socks5 / http / https doesn't have any authentication mechanism.

zonyitoo:
[...]

Issue #232, opened 2020-05-01
Shadowsocks should send data along with header


Mygod:
Currently, this implementation does handshake first, which is problematic for obfuscation, e.g.

https://github.com/shadowsocks/shadowsocks-rust/blob/1204d581610c5c6bcb24aab40e1d4fa52c18aa29/src/relay/tcprelay/socks5_local.rs#L40

zonyitoo:
Hmm, IV is already sent with the first packet:

1. Client -> Server: target's `Address`
2. Server -> Client: the first response packet.

Mygod:
Two connections should use independent IVs.

zonyitoo:
Yes, they are.

Mygod:
The first packet should contain the header/handshake as well as actual payload.

zonyitoo:
Hmm, then the `ProxyStream` should store the `Address` and sent it with the first `write` call.

Mygod:
CC @madeye to double check on this.

zonyitoo:
Made a simple but a little bit ugly solution.

Mygod:
This should also be done at the server side as well -- https://github.com/shadowsocks/shadowsocks-rust/blob/806f44244af1366323d69a2abed3a530838ab264/src/relay/tcprelay/aead.rs

zonyitoo:
Server side doesn't have a handshake packet.

Mygod:
Cool just double checked via `tcpdump` and can confirm that both sides are doing it correctly with https://github.com/shadowsocks/shadowsocks-rust/commit/45a3469840fa423cecec90de5cc03ed5003c1a92.

Mygod:
@madeye Maybe we should include this in official doc as well?

madeye:
Yeah, we should do that.

Mygod:
@madeye Updated at https://github.com/shadowsocks/shadowsocks-org/wiki/Protocol

madeye:
Hmm, this fix caused another problem: https://github.com/shadowsocks/shadowsocks-android/issues/2621.

We'd better have a delay to send out the header, even if there's no data. Some underlying protocol requires this behavior like TCP.

zonyitoo:
Well, it would be quite tricky to do that in a `Future` based server... But it should be able to do.

madeye:
Yes, I tried to implement this myself at first, but found it's not easy.

zonyitoo:
[...]

PR #345, opened 2020-12-13
Refactored and separate library into crates

### LINKED ISSUE #343: timeout handling is completely broken
Hello,
Shadowsocks applies the timeout not on receiving first data from the client, but on any traffic activity in the entire connection. This is completely wrong and has been fixed before in shadowsocks-libev a very long time ago (years). Even Cloak had this problem and I had to fix it myself.

Why this is wrong ?
Because shadowsocks is not a sockets application, it is just a proxy. A proxy should not assume ownership and enforce its own timeouts. It should be completely transparent and not alter the characteristics of the connection. It is up to the Layer 8 (?) application that's running on TOP of shadowsocks to implement its own timeouts or keepalives in the application layer (eg: SSH sending of null packets, HTTP keep alives, HTTP timeouts), and not the job of a proxy application.

The only kind of timeout that's permissible for a "proxy" kind of application to enforce is a timeout for the **initial** data reception from the local application, or from the client (in the server's point of view) and ONLY as a kind of defense from resource exhaustion attacks.

Otherwise, many kinds of problems will happen to proxied applications (eg: killing of Idle SSH sessions) or termination of long idle connections that are supposed to be completely valid.

I keep seeing this mistake being repeated in application after application. Please fix this. I don't know rust to fix this myself and now shadowsocks-libev project is directing users to use this one instead which has this serious regression of behavior.

Thank you for understanding.

zonyitoo:
- [x] shadowsocks: the core shadowsocks library
- [x] shadowsocks-service:
    - [x] server
    - [x] manager
    - [x] local
        - [x] socks5, socks4/4a
        - [x] tunnel
        - [x] http/https
        - [x] redir
        - [x] dns
- [x] shadowsocks-rust: release binaries
- [x] tests

ref #326.

This PR will also fixes #343 , #232 .


zonyitoo:
@madeye @Mygod This branch is just finished. Please check if it is compatible with shadowsocks-android.

madeye:
Several issues observed now:

1. crash on x86 emulator:

```
12-18 09:09:32.090 10829 10829 I crash_dump32: performing dump of process 10824 (target tid = 10824)
12-18 09:09:32.090 10829 10829 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
12-18 09:09:32.091 10829 10829 F DEBUG   : Build fingerprint: 'google/sdk_gphone_x86/generic_x86:10/QSR1.200715.002/6695061:userdebug/dev-keys'
12-18 09:09:32.091 10829 10829 F DEBUG   : Revision: '0'
12-18 09:09:32.091 10829 10829 F DEBUG   : ABI: 'x86'
12-18 09:09:32.093 10829 10829 F DEBUG   : Timestamp: 2020-12-18 09:09:32+0800
12-18 09:09:32.093 10829 10829 F DEBUG   : pid: 10824, tid: 10824, name: libsslocal.so  >>> /data/app/com.github.shadowsocks-h7SbUFhjh9RgI6i9YGsIVQ==/lib/x86/libsslocal.so <<<
12-18 09:09:32.093 10829 10829 F DEBUG   : uid: 10133
12-18 09:09:32.093 10829 10829 F DEBUG   : signal 31 (SIGSYS), code 1 (SYS_SECCOMP), fault addr --------
12-18 09:09:32.093 10829 10829 F DEBUG   : Cause: seccomp prevented call to disallowed x86 system call 364
```

The latest binary calls "x86 system call 364", which is allowed on Android 10 and above.
With some study, it looks 364 syscall on x86 is accept4 defined here: https://elixir.bootlin.com/linux/v5.6/source/net/socket.c#L1820

Some references: https://www.lurklurk.org/syscalls.html

2. UDP only mode reports:

```
config integrity check failed, socks protocol must enable tcp mode
```

3. Minor fixes for android target:

```diff
diff --git a/bin/sslocal.rs b/bin/sslocal.rs
index 10443d8..0990a73 100644
--- a/bin/sslocal.rs
+++ b/bin/sslocal.rs
@@ -249,7 +249,7 @@ fn main() {
     if matches.is_present("VPN_MODE") {
         // A socket `protect_path` in CWD
         // Same as shadowsocks-libev's android.c
-        config.protect_path = Some(From::from("protect_path"));
+        config.outbound_vpn_protect_path = Some(From::from("protect_path"));

         // Set default config.local_dns_addr
         #[cfg(feature = "local-dns")]
```

4. MISC warnings:

```
warning: unused `AsyncFdReadyGuard` that must be used
  --> crates/shadowsocks/src/relay/sys/unix/uds.rs:31:9
   |
31 |         future::poll_fn(|cx| io.poll_write_ready(cx)).await?;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_must_use)]` on by default
   = note: You must explicitly choose whether to clear the readiness state by calling a method on ReadyGuard

warning: unused `AsyncFdReadyGuard` that must be used
  --> crates/shadowsocks/src/relay/sys/unix/uds.rs:36:9
   |
36 |         ready!(self.io.poll_write_ready(cx))?;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: You must explicitly choose whether to clear the readiness state by calling a method on ReadyGuard

warning: 2 warnings emitted

warning: unused variable: `nofile`
  --> crates/shadowsocks-service/src/sys/unix/mod.rs:44:19
   |
44 | pub fn set_nofile(nofile: u64) -> io::Result<()> {
   |                   ^^^^^^ help: if this is intentional, prefix it with an underscore: `_nofile`
   |
   = note: `#[warn(unused_variables)]` on by default

warning: 1 warning emitted
```

madeye:
The `local-dns` feature seems not working on Android.

Is the previous `local-dns-relay` feature through UNIX domain socket still supported?

zonyitoo:
> The `local-dns` feature seems not working on Android.
> 
> Is the previous `local-dns-relay` feature through UNIX domain socket still supported?

Yes, it is supported. But I couldn't verify that locally.

zonyitoo:
> config integrity check failed, socks protocol must enable tcp mode

Maybe socks5 shouldn't support `udp_only` mode? Because UDP association command requires tcp handshake.

madeye:
I found the local DNS server only listening on TCP ports. Is it possible to listen on UDP DNS port but forward the DNS queries to local UNIX domain socket?

zonyitoo:
It should works..

https://github.com/shadowsocks/shadowsocks-rust/blob/5fb8b31fc9be11f1bf7d851353dd70320102ab70/crates/shadowsocks-service/src/local/dns/server.rs#L665-L671

Currently the DNS relay server also uses `mode` to decide which protocol it is going to use. So if `mode` is `tcp_only`, then only the TCP port is listened.

Should I ignore the `mode` parameter and listen to both TCP and UDP port?

zonyitoo:
> The latest binary calls "x86 system call 364", which is not allowed on Android 10 and above.

Doomed. https://github.com/tokio-rs/mio/blob/290c43a96662d54ab7c4b8814e5a9f9a9e523fda/src/sys/unix/tcp.rs#L435-L440

madeye:
> Should I ignore the mode parameter and listen to both TCP and UDP port?

Yes, I think so.

> Because the current master branch is also using tokio 0.3, which is already depends on mio 0.7 (including accept4).

Yes, the current master branch has the same issue...

I didn't observe this on any real Android devices. Maybe it's just an issue for Emulator.

zonyitoo:
The latest commit fixed those issues except `accept4`.

madeye:
The `lookup_tcp_remote` seems not protected by android socket protector.

I see no successful DNS resolves.

zonyitoo:
> The `lookup_tcp_remote` seems not protected by android socket protector.
> 
> I see no successful DNS resolves.

The latest commit fixed bug while handling the `mode` in DNS relay. Now `mode` is for controlling upstreams, connect remote with TCP or UDP, or both.

`lookup_tcp_remote` should work the same because it uses the same `ConnectOpts` structure. Could you provide more detail about it ? (I can't reproduce it because I don't have an Android device, `protect_path` never runs :( )

madeye:
Thanks for the info. Let me debug it locally first.

zonyitoo:
You may try to add a `trace!` log in that function to print out the `ConnectOpts` structure and see if its `vpn_protect_path` is set properly.

madeye:
Okay, now TCP DNS queries will reuse the connections. However, our current Android local resolver only process one query each connection. I'll try to fix it.

BTW, what if a TCP DNS connection is closed? Do we have the logic to reconnect?

zonyitoo:
Well, the socket taken from cache will be dropped if request failed. But that client request is also failed :P .

zonyitoo:
With this commit, it should work even if you have closed connections.

madeye:
BTW, when forwarding a UDP DNS query to TCP upstream, have we added the length before the query?

I'm trying get sslocal works on my desktop with TCP only mode sslocal plus UDP only local DNS resolver. However, I cannot get a valid response.

zonyitoo:
It works well on my laptop. And yes, I have added the length before the query.

madeye:
Yes, it's my problem. I forgot to add server port to the local-dns-addr. Now it works.

madeye:
Okay, it looks related to plugin. I cannot get TCP stream works with plugins...

zonyitoo:
Have plugin started?

madeye:
Yes.

It looks I can reproduce this on my desktop. Can you confirm the plugin feature works now?

madeye:
Hmm. I found the problem. The rc4-md5 cipher is broken now...

database64128:
> Hmm. I found the problem. The rc4-md5 cipher is broken now...

How about we take the chance and drop support for all stream ciphers once and for all? 😁

zonyitoo:
> Hmm. I found the problem. The rc4-md5 cipher is broken now...

@LuoZijun Please take a look at this issue. rc4-md5 compatibility issue.

madeye:
> Maybe socks5 shouldn't support udp_only mode? 

Shadowsocks-android has a UDP fallback feature, which starts a UDP only socks5 client with `--udp-bind-addr` option.

zonyitoo:
> Hmm. I found the problem. The rc4-md5 cipher is broken now...

I don't think so. I tested with `libev`'s `ss-local` and `ss-server`, all of them works perfectly with `"method": "rc4-md5"`

madeye:
> I don't think so. I tested with libev's ss-local and ss-server, all of them works perfectly with "method": "rc4-md5"

Interesting, I cannot get it work locally. I'm testing on a Ubuntu x86_64 machine.

madeye:
Here is the config file I am using now:

```
{
    "server":"127.0.0.1",
    "mode":"tcp_and_udp",
    "server_port":8388,
    "local_port":1080,
    "password":"barfoo!",
    "timeout":86400,
    "method":"rc4-md5"
}
```

madeye:
I tried updating my nightly rust toolchains, and rc4-md5 works now...

madeye:
It looks there's a typo causing old DNS resolver in TCP_ONLY mode.

```diff
diff --git a/crates/shadowsocks-service/src/local/dns/server.rs b/crates/shadowsocks-service/src/local/dns/server.rs
index 1699f54..d89ff50 100644
--- a/crates/shadowsocks-service/src/local/dns/server.rs
+++ b/crates/shadowsocks-service/src/local/dns/server.rs
@@ -469,7 +469,7 @@ impl DnsClient {
             None
         };

-        let udp_balancer = if mode.enable_tcp() {
+        let udp_balancer = if mode.enable_udp() {
             let mut balancer_builder = PingBalancerBuilder::new(context.clone(), BalancerServerType::Udp);

             for server in servers {
```

zonyitoo:
> ~I tried updating my nightly rust toolchains, and rc4-md5 works now...~
> 
> EDIT: Using the wrong config.json, rc4-md5 still doesn't work...

Yes, I can confirm that it is not working.

zonyitoo:
Bug fixed. A stupid bug.

zonyitoo:
# Naive Performance Test
## Client and Target Server configuration

* Target Server:

```bash
$ iperf -s
------------------------------------------------------------
Server listening on TCP port 5001
TCP window size:  128 KByte (default)
------------------------------------------------------------
```

* Test Client:

```bash
$ iperf -c 127.0.0.1 -p 1300
```

## Stream Test

* method: `rc4-md5`

|                    | ss-local (libev) | sslocal (rust) |
| ------------------ | ---------------- | -------------- |
| ss-server (libev)  | 4.09 GBytes  3.51 Gbits/sec | 3.34 GBytes  2.87 Gbits/sec |
| ssserver (rust)    | 3.41 GBytes  2.93 Gbits/sec | 3.41 GBytes  2.93 Gbits/sec |

## AEAD Test

* method: `aes-128-gcm`

|                    | ss-local (libev) | sslocal (rust) |
| ------------------ | ---------------- | -------------- |
| ss-server (libev)  | 2.12 GBytes  1.82 Gbits/sec | 2.06 GBytes  1.77 Gbits/sec |
| ssserver (rust)    | 2.18 GBytes  1.87 Gbits/sec | 9.11 GBytes  7.82 Gbits/sec |

## Result

* Stream mode has significant performance issue
* AEAD mode is blazingly faster than libev

zonyitoo:
## Stream Test 2

* method: `aes-128-cfb`

|                    | ss-local (libev) | sslocal (rust) |
| ------------------ | ---------------- | -------------- |
| ss-server (libev)  | 3.52 GBytes  3.02 Gbits/sec | 3.56 GBytes  3.06 Gbits/sec |
| ssserver (rust)    | 3.60 GBytes  3.09 Gbits/sec | 3.58 GBytes  3.08 Gbits/sec |

Another test shows different result, `rust` and `libev` doesn't have any significant differences.

madeye:
Some build issues on Android:

```diff
diff --git a/crates/shadowsocks-service/src/config.rs b/crates/shadowsocks-service/src/config.rs
index 58ea7c2..bde46f6 100644
--- a/crates/shadowsocks-service/src/config.rs
+++ b/crates/shadowsocks-service/src/config.rs
@@ -59,7 +59,7 @@ use std::{

 use cfg_if::cfg_if;
 use serde::{Deserialize, Serialize};
-#[cfg(feature = "local-tunnel")]
+#[cfg(any(feature = "local-tunnel", feature = "local-dns"))]
 use shadowsocks::relay::socks5::Address;
 use shadowsocks::{
     config::{ManagerAddr, ServerAddr, ServerConfig},
diff --git a/crates/shadowsocks-service/src/manager/mod.rs b/crates/shadowsocks-service/src/manager/mod.rs
index c85c04f..8577b73 100644
--- a/crates/shadowsocks-service/src/manager/mod.rs
+++ b/crates/shadowsocks-service/src/manager/mod.rs
@@ -2,10 +2,9 @@
 //!
 //! Service for managing multiple relay servers. [Manage Multiple Users](https://github.com/shadowsocks/shadowsocks/wiki/Manage-Multiple-Users)

-use std::{
-    io::{self, ErrorKind},
-    sync::Arc,
-};
+use std::io::{self, ErrorKind};
+#[cfg(feature = "trust-dns")]
+use std::sync::Arc;

 use log::{trace, warn};
 use shadowsocks::{config::ServerAddr, net::ConnectOpts};
diff --git a/crates/shadowsocks-service/src/server/mod.rs b/crates/shadowsocks-service/src/server/mod.rs
index 34a9670..726b182 100644
--- a/crates/shadowsocks-service/src/server/mod.rs
+++ b/crates/shadowsocks-service/src/server/mod.rs
@@ -7,7 +7,9 @@ use std::{

 use futures::{future, FutureExt};
 use log::{trace, warn};
-use shadowsocks::{config::ServerAddr, dns_resolver::DnsResolver, net::ConnectOpts};
+#[cfg(feature = "trust-dns")]
+use shadowsocks::dns_resolver::DnsResolver;
+use shadowsocks::{config::ServerAddr, net::ConnectOpts};

 use crate::config::{Config, ConfigType};
```

madeye:
Okay, the stream cipher fix is verified.

So, the only issue left is how to handle UDP only socks5 association mode, which is required for Android.

zonyitoo:
> Okay, the stream cipher fix is verified.
> 
> So, the only issue left is how to handle UDP only socks5 association mode, which is required for Android.

It should already be fixed, too.

madeye:
Still see this error on Android:

https://github.com/shadowsocks/shadowsocks-rust/blob/0f67279369040030f570a845acf8e8b60f6ed044/crates/shadowsocks-service/src/config.rs#L1187

madeye:
Okay, I found a way to fix this from Android side.

Currently, if the server doesn't support UDP relay, but the client enables it, we would see extra latency for local-dns resolver. It's because shadowsocks-rust always sends DNS requests through UDP first and wait for timeout.

We may add a heuristic feature here to disable UDP DNS queries if the UDP request are always timeout.

madeye:
UDP relay became Symmetric cone now (NAT3). To achieve full cone (NAT2), we should not include `dst addr` in the key of LRU cache: 

https://github.com/shadowsocks/shadowsocks-rust/blob/0f67279369040030f570a845acf8e8b60f6ed044/crates/shadowsocks-service/src/local/socks/server/socks5/udprelay.rs#L54

In shadowsocks-libev, we only use source address to generate the keys:

https://github.com/shadowsocks/shadowsocks-libev/blob/aed81a1c2b44505563c5c969b4546bebd4745dde/src/udprelay.c#L627

zonyitoo:
> Okay, I found a way to fix this from Android side.

Could you elaborate more about Android's UDP fallback mode? Why would Android need a socks5 server in `udp_mode`?

> Currently, if the server doesn't support UDP relay, but the client enables it, we would see extra latency for local-dns resolver. It's because shadowsocks-rust always sends DNS requests through UDP first and wait for timeout.
> 
> We may add a heuristic feature here to disable UDP DNS queries if the UDP request are always timeout.

Sounds reasonable. But this feature could be controlled by `"mode": "tcp_only"`.

1. If socks5 server enables UDP mode, then it implies that remote server supports UDP relay.
2. If DNS relay is running in standalone mode, then you can also control this behavior with `mode`

madeye:
> Or maybe, I should send 2 queries simutanously?

Sounds risky. because it will become a pattern of shadowsocks client.

@Mygod  What about add a TCP only mode option to shadowsocks-android.

> Could you elaborate more about Android's UDP fallback mode? Why would Android need a socks5 server in udp_mode?

We have some profiles works in TCP only mode due to the SIP003 plugin. So we added a UDP fallback model to let user route UDP traffic to some other profiles.

madeye:
@zonyitoo I still see `sslocal` listening on TCP port even in UDP only mode:

```
./sslocal -s test.com:8038 -k test -m chacha20-ietf-poly1305 -b 127.0.0.1:1080 -u
2020-12-19T18:24:09.429522500+08:00 INFO  shadowsocks 1.9.0
2020-12-19T18:24:09.431519100+08:00 INFO  shadowsocks socks TCP listening on 127.0.0.1:1080
2020-12-19T18:24:09.431862+08:00 INFO  shadowsocks socks5 UDP listening on 127.0.0.1:1080
```

LuoZijun:
> @zonyitoo I still see `sslocal` listening on TCP port even in UDP only mode:
> 

`SOCKS5_CMD_UDP_ASSOCIATE` 需要通过 TCP 协议协商建立，所以TCP协议是必须的，通过 TCP 通道向  sslocal 的 SOCKS5 Server 请求 UDP 通道。

madeye:
> 从 SOCKS5 协议的机制上面就无法支持 UDP-Only ，所以确定的说只有2种工作模式：TCP-Only 和 TCP && UDP 。

We're not talking about the standard SOCKS5 protocol. The "UDP only socks5" is just a special mode for shadowsocks-android, which aims to provide a UDP only socks5 service to a different upstream.

@zonyitoo I just found a solution with `--udp-bind-address`, so please ignore the previous request for now.

madeye:
Here is a fix for UDP relay (I only fixed the local socks5 client, similar fix should be applied to local redir and server sides):

```diff
--- a/crates/shadowsocks-service/src/local/socks/server/socks5/udprelay.rs
+++ b/crates/shadowsocks-service/src/local/socks/server/socks5/udprelay.rs
@@ -51,7 +51,7 @@ struct Socks5Bound {
 
 pub struct Socks5UdpServer {
     context: Arc<ServiceContext>,
-    assoc_map: Arc<Mutex<LruCache<Socks5Bound, UdpAssociation>>>,
+    assoc_map: Arc<Mutex<LruCache<SocketAddr, UdpAssociation>>>,
 }
 
 impl Socks5UdpServer {
@@ -154,7 +154,7 @@ impl Socks5UdpServer {
         data: &[u8],
     ) -> io::Result<()> {
         let mut assoc_map = self.assoc_map.lock().await;
-        let assoc = match assoc_map.entry(assoc_key.clone()) {
+        let assoc = match assoc_map.entry(assoc_key.src.clone()) {
             Entry::Occupied(occ) => occ.into_mut(),
             Entry::Vacant(vac) => {
                 let target_addr = &assoc_key.dst;
@@ -283,15 +283,15 @@ impl UdpAssociation {
         inbound: Arc<UdpSocket>,
         assoc_key: Socks5Bound,
         outbound: Arc<MonProxySocket>,
-        assoc_map: Arc<Mutex<LruCache<Socks5Bound, UdpAssociation>>>,
+        assoc_map: Arc<Mutex<LruCache<SocketAddr, UdpAssociation>>>,
     ) -> io::Result<()> {
         let mut buffer = [0u8; MAXIMUM_UDP_PAYLOAD_SIZE];
         let mut payload_buffer = BytesMut::new();
         loop {
-            let (n, _) = match outbound.recv(&mut buffer).await {
+            let (n, addr) = match outbound.recv(&mut buffer).await {
                 Ok(n) => {
                     // Keep association alive in map
-                    let _ = assoc_map.lock().await.get(&assoc_key);
+                    let _ = assoc_map.lock().await.get(&assoc_key.src);
                     n
                 }
                 Err(err) => {
@@ -308,7 +308,7 @@ impl UdpAssociation {
             payload_buffer.clear();
 
             // Resssemble packet
-            let header = UdpAssociateHeader::new(0, assoc_key.dst.clone());
+            let header = UdpAssociateHeader::new(0, addr.clone());
             payload_buffer.reserve(header.serialized_len() + n);
 
             header.write_to_buf(&mut payload_buffer);
@@ -353,14 +353,14 @@ impl UdpAssociation {
         inbound: Arc<UdpSocket>,
         assoc_key: Socks5Bound,
         outbound: Arc<UdpSocket>,
-        assoc_map: Arc<Mutex<LruCache<Socks5Bound, UdpAssociation>>>,
+        assoc_map: Arc<Mutex<LruCache<SocketAddr, UdpAssociation>>>,
     ) -> io::Result<()> {
         let mut buffer = [0u8; MAXIMUM_UDP_PAYLOAD_SIZE];
         loop {
             let n = match outbound.recv(&mut buffer).await {
                 Ok(n) => {
                     // Keep association alive in map
-                    let _ = assoc_map.lock().await.get(&assoc_key);
+                    let _ = assoc_map.lock().await.get(&assoc_key.src);
                     n
                 }
                 Err(err) => {
```

zonyitoo:
> @zonyitoo I still see `sslocal` listening on TCP port even in UDP only mode:
> 
> ```
> ./sslocal -s test.com:8038 -k test -m chacha20-ietf-poly1305 -b 127.0.0.1:1080 -u
> 2020-12-19T18:24:09.429522500+08:00 INFO  shadowsocks 1.9.0
> 2020-12-19T18:24:09.431519100+08:00 INFO  shadowsocks socks TCP listening on 127.0.0.1:1080
> 2020-12-19T18:24:09.431862+08:00 INFO  shadowsocks socks5 UDP listening on 127.0.0.1:1080
> ```

Well, socks5's UDP associate command requires TCP handshake.

Or ... I should completely disable TCP listener?

zonyitoo:
> Here is a fix for UDP relay (I only fixed the local socks5 client, similar fix should be applied to local redir and server sides):

I am working on it. It is not a simple task, because one client may send to different targets, including bypass and proxied. So I have to dynamically create bypass task and proxy task for one client.

zonyitoo:
> I just found a solution with `--udp-bind-address`, so please ignore the previous request for now.

What kind of solution?

madeye:
I just start the UDP fallback sslocal on another port, and set udp-bind-addr of the TCP only sslocal to the fall back one.

It would be very helpful to not listen on TCP port when UDP only mode is enabled.

zonyitoo:
> It would be very helpful to not listen on TCP port when UDP only mode is enabled.

Working on it.

zonyitoo:
`udp_max_associations` won't have a default value, which have to be set explictly.

madeye:
I think this PR is almost ready for shadowsocks-android.

BTW, still see one warning:

```
warning: unused import: `ErrorKind`  --> crates/shadowsocks-service/src/local/socks/server/mod.rs:4:16 
```

zonyitoo:
There still have some dependencies that are migrating to tokio v0.3 are not released.

cg31:
This is a good move. But how am I supposed to use the crates from other project?

It seems I can't refer to both shadowsocks and shadowsocks-service on github:
```
[dependencies]
shadowsocks = { git = "https://github.com/shadowsocks/shadowsocks-rust.git", branch = "master" }
# or ???
shadowsocks-server = { git = "https://github.com/shadowsocks/shadowsocks-rust.git", branch = "master" }

````
I can only clone them to local and use them:
```
[dependencies]
shadowsocks = { path = "./shadowsocks/crates/shadowsocks" }
shadowsocks-service = { path = "./shadowsocks/crates/shadowsocks-service" }
```
but this is rather awkward.

**Comment by zonyitoo** [comment]
```toml
[dependencies]
shadowsocks = { git = "https://github.com/shadowsocks/shadowsocks-rust.git", branch = "master" }
shadowsocks-service = { git = "https://github.com/shadowsocks/shadowsocks-rust.git", branch = "master" }
```

It should work.

Mygod:
What is `DnsClientCache`? Caching should be handled by shadowsocks-android since shadowsocks-rust is oblivious to underlying network changes.

zonyitoo:
> What is `DnsClientCache`? Caching should be handled by shadowsocks-android since shadowsocks-rust is oblivious to underlying network changes.

Just for preventing creating sockets for every queries.

If the underlying network changed, which will definitely cause query failure, then the `DnsClientCache` will automatically recreate sockets and do it again.

Mygod:
Does this fix https://github.com/shadowsocks/shadowsocks-android/issues/2093?

Mygod:
Also according to [rfc7766](https://datatracker.ietf.org/doc/html/rfc7766), connection reuse can send multiple queries in a single connection. 🤔

zonyitoo:
> Does this fix [shadowsocks/shadowsocks-android#2093](https://github.com/shadowsocks/shadowsocks-android/issues/2093)?

Well, the `DnsClientCache` should achieve the goal of DNS connection reuse (TCP).

Issue #394, opened 2021-01-24
[Feature proposal] Server disable in Extended multiple server configuration


spyophobia:
## Rationale
`shadowsocks-rust` supports creating and/or connecting to multiple servers within one configuration file, which is great.

However, sometimes there arises the need, whether during testing or deployment, for us to temporarily disable one or more servers. Example scenarios include:
- Several conflicting server profiles during testing
- Isolating the issue during network diagnosis
- Server temporarily unavailable

Ideally, the user should be able to simply block-comment out a profile to disable it. Unfortunately, JSON format does not have support for comments, which means the user would need to:
1. either joggle several configuration files
2. or copying & pasting the profile around in the JSON file

... Neither of which is convenient.

## What I am envisioning
Wouldn't it be nice, if you can simply add `"disable": true` to a server profile and turn it off?

Something like this:
```json
{
    "servers": [
        {
            "address": "127.0.0.1",
            "port": 1080,
            "password": "hello-world",
            "method": "aes-256-gcm",
            "timeout": 300
        },
        {
            "disable": true,
            "address": "127.0.0.1",
            "port": 1081,
            "password": "hello-kitty",
            "method": "chacha20-ietf-poly1305"
        }
    ],
    "local_port": 8388,
    "local_address": "127.0.0.1"
}
```

I think it would be a good "quality of life" improvement while being not too complicated to implement.

spyophobia:
Also, I personally know some Rust so I'm happy to contribute via a PR.

spyophobia:
Also, I personally know some Rust so I'm happy to contribute via a PR.

database64128:
+1. The feature would be helpful for me as well.

zgzh:
If possible, set the specified server as the default, and use other servers when it fails. This scenario is that the network speed is fast, but the speed is low, so it can’t be optimized. You can only delete the server.

spyophobia:
> If possible, set the specified server as the default, and use other servers when it fails. This scenario is that the network speed is fast, but the speed is low, so it can’t be optimized. You can only delete the server.

Don't worry I got you covered. I actually intend to submit another feature proposal about customisation options for the load balancer, which will be a superset of the feature you are describing.

spyophobia:
@zgzh See https://github.com/shadowsocks/shadowsocks-rust/issues/395

zonyitoo:
[...]

PR #401, opened 2021-01-27
Lazy server disable implementation

spyophobia:
See shadowsocks/shadowsocks-rust#394

This is a lazy implementation that simply filters for `disable` before the vector of `shadowsocks::ServerConfig` is created.

- Benefit: simply, lazy, no memory usage, no modification to `shadowsocks::ServerConfig`
- Downside: `Display` trait for `shadowsocks_service::Config` is not aware of disabled servers

### DIFF
--- a/crates/shadowsocks-service/src/config.rs
+++ b/crates/shadowsocks-service/src/config.rs
@@ -138,6 +138,8 @@ struct SSServerExtConfig {
     password: String,
     method: String,
     #[serde(skip_serializing_if = "Option::is_none")]
+    disable: Option<bool>,
+    #[serde(skip_serializing_if = "Option::is_none")]
     plugin: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     plugin_opts: Option<String>,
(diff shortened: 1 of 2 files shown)


crates/shadowsocks-service/src/config.rs
```diff
@@ -872,7 +874,16 @@ impl Config {
...
+            // Filter out disabled servers
+            let enabled_servers = servers
+                .into_iter()
+                .filter(|ext_cfg| match ext_cfg.disable {
+                    Some(true) => false,
+                    _ => true,
+                })
+                .collect::<Vec<_>>();
+
+            for svr in enabled_servers {
```
zonyitoo, line 886: This `collect` is completely unnecessary. Just do

```rust
for svr in servers {
    if svr.disable.unwrap_or(false) {
        continue;
    }
}
```

zonyitoo (changes requested):

crates/shadowsocks-service/src/config.rs
spyophobia: Sure

spyophobia (commented):
Done

PR #435, opened 2021-03-01
Update debian/changelog (v1.9.1)

kimw:


zonyitoo:
How to genereate this file from release?

kimw:
It's "generated" by my bare hands 🥲

---

1. Copy change log content from release info.
2. Catch date time from git log by command such as `git show --date=rfc v1.9.1`

Barely, huh?

kimw:
@zonyitoo how do you make the release? Maybe we can,

1. Write release info into `debian/changelog`
2. Run a script to release to GitHub with the information from `debian/changelog`

How's about it?

zonyitoo:
Currently there is another PR for building RPM #436 . So .. maybe we should have a common `CHANGELOG` file and then generates it to different targets.

PR #450, opened 2021-03-10
add release i386

1715173329:
This adds i386 builds in workflow.

BTW, I'm confused with the difference between `arm-unknown-linux-musleabihf` and
`armv7-unknown-linux-musleabihf`, would anyone tell me that? Thanks so much.

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -15,9 +15,10 @@ jobs:
     strategy:
       matrix:
         target:
+          - i686-unknown-linux-musl
           - x86_64-pc-windows-gnu
-          - x86_64-unknown-linux-musl
           - x86_64-unknown-linux-gnu
+          - x86_64-unknown-linux-musl
           - arm-unknown-linux-gnueabi
           - arm-unknown-linux-gnueabihf
           - arm-unknown-linux-musleabi


zonyitoo:
I don't know about it either. But why adding so many i386 targets? Do anyone still using windows running on i386?

1715173329:
> I don't know about it either. 

Okay, thanks for replying.

> But why adding so many i386 targets? Do anyone still using windows running on i386?

Yes, there's still some users use i386 based device as their routers (linux, musl)...
And like me, uses windows running on i386, but it's not really necessary, I can drop this target.

zonyitoo:
Yes please drop it. I don't think it should be used right now. `i686-pc-windows-gnu`

And also, we can still find CPUs that are i586 targets?

1715173329:
> Yes please drop it. I don't think it should be used right now. `i686-pc-windows-gnu`

All right.

> And also, we can still find CPUs that are i586 targets?

In my option it's really a bad idea that running any modern softwares as well as systems
on i586 target. But I can never know what they thought, like:
AdguardTeam/AdGuardHome#996

zonyitoo:
i586     –     Intel Pentium         (in 1993)          or          AMD-K5 (in 1996)

I don't think we should waste any release time on these platforms.

1715173329:
> I don't think we should waste any release time on these platforms.

The only reason I added it here is there exists such a target on OpenWrt,
for nothing else. They're even much older than me 🤣🤣

Okay I dropped them.

zonyitoo:
Well, do you need those i686 linux targets?

1715173329:
> Well, do you need those i686 linux targets?

For musl only, I can drop gnu one.

PR #556, opened 2021-06-19
Remove the behavior of silent dropping replay packets

Mygod:
This removes a carefully engineered feature (proposed in https://github.com/shadowsocks/shadowsocks-org/issues/44#issuecomment-280007952) to detect Shadowsocks as per https://github.com/net4people/bbs/issues/22. Testing in progress.

### DIFF
--- a/crates/shadowsocks/src/relay/tcprelay/aead.rs
+++ b/crates/shadowsocks/src/relay/tcprelay/aead.rs
@@ -43,7 +43,7 @@ use std::{
 use byte_string::ByteStr;
 use bytes::{BufMut, Bytes, BytesMut};
 use futures::ready;
-use log::trace;
+use log::{trace, warn};
 use tokio::io::{AsyncRead, AsyncWrite, ReadBuf};
 
 use crate::{
@@ -223,12 +223,7 @@ impl DecryptedReader {
             let salt = self.salt.take().unwrap();
 
             if context.check_nonce_and_set(&salt) {
-                use std::io::Error;
-
-                trace!("detected repeated AEAD salt {:?}", ByteStr::new(&salt));
-
-                let err = Error::new(ErrorKind::Other, "detected repeated salt");
-                return Err(err).into();
+                warn!("detected repeated AEAD salt {:?}", ByteStr::new(&salt));
             }
         }
 
--- a/crates/shadowsocks/src/relay/tcprelay/stream.rs
+++ b/crates/shadowsocks/src/relay/tcprelay/stream.rs
@@ -10,7 +10,7 @@ use std::{
 use byte_string::ByteStr;
 use bytes::{BufMut, Bytes, BytesMut};
 use futures::ready;
-use log::trace;
+use log::{trace, warn};
 use tokio::io::{AsyncRead, AsyncWrite, ReadBuf};
 
 use crate::{
@@ -113,12 +113,7 @@ impl DecryptedReader {
 
         let iv = &self.buffer[..iv_len];
         if context.check_nonce_and_set(&iv) {
-            use std::io::Error;
-
-            trace!("detected repeated stream iv {:?}", ByteStr::new(&iv));
-
-            let err = Error::new(ErrorKind::Other, "detected repeated iv");
-            return Err(err).into();
+            warn!("detected repeated stream iv {:?}", ByteStr::new(&iv));
         }
 
         trace!("got stream iv {:?}", ByteStr::new(iv));
--- a/crates/shadowsocks/src/relay/udprelay/crypto_io.rs
+++ b/crates/shadowsocks/src/relay/udprelay/crypto_io.rs
@@ -23,7 +23,7 @@ use std::io::{self, Cursor, ErrorKind};
 
 use byte_string::ByteStr;
 use bytes::{BufMut, BytesMut};
-use log::{debug, trace};
+use log::{trace, warn};
 
 use crate::{
     context::Context,
@@ -182,8 +182,7 @@ async fn decrypt_payload_stream(
 
     let (iv, data) = payload.split_at_mut(iv_len);
     if context.check_nonce_and_set(iv) {
-        debug!("detected repeated iv {:?}", ByteStr::new(iv));
-        return Err(io::Error::new(io::ErrorKind::Other, "detected repeated iv"));
+        warn!("detected repeated iv {:?}", ByteStr::new(iv));
     }
 
     trace!("UDP packet got stream IV {:?}", ByteStr::new(iv));
@@ -215,8 +214,7 @@ async fn decrypt_payload_aead(
 
     let (salt, data) = payload.split_at_mut(salt_len);
     if context.check_nonce_and_set(salt) {
-        debug!("detected repeated salt {:?}", ByteStr::new(salt));
-        return Err(io::Error::new(io::ErrorKind::Other, "detected repeated salt"));
+        warn!("detected repeated salt {:?}", ByteStr::new(salt));
     }
 
     trace!("UDP packet got AEAD salt {:?}", ByteStr::new(salt));


Mygod:
CC @madeye Feel free to test.

zonyitoo:
Hmm, but why removing it?

Mygod:
Because Bloom filter itself is a feature that can be utilized by GFW. See the post I linked above.

database64128:
The report you linked as a reference contradicts your point. Shadowsocks servers **need** protection against replay attacks. The current bloomfilter-based implementation might not be good enough (limited capacity, no persistency, etc). But removing the protection is not the solution. We need protocol changes to achieve full protection.

Since you have consistently given downvotes/objections to proposals of any change (shadowsocks/shadowsocks-org#177, shadowsocks/shadowsocks-org#178, shadowsocks/shadowsocks-org#183 (comments)), I'm giving your PR a 👎.

zonyitoo:
Since there is no better solution for now, doing nothing should be better than removing it completely. https://github.com/net4people/bbs/issues/22 actually indicates that the main detection method is **replay**.

Mygod:
Let me clarify by giving an explicit attack to distinguish/detect shadowsocks traffic. Most protocols that looks random allow replay of the first packet. The only counterexample I can think of is TLS 1.3 0-RTT, but even there, the behavior is immediately different. shadowsocks-libev and shadowsocks-rust disallows replay within a certain period (until server restarts or ping-pong bloom filter resets). This by itself is a feature that an active probing adversary can utilize to distinguish shadowsocks-rust from TLS by doing replay. To perform this distinguishing attack, you need:

1. Immediate replay of the packet, and verify that the connection is silently dropped.
2. Replay of the packet after a random amount of time, and verify that the server eventually accepts and returns some data.
3. Replay of the packet (also after a random amount of time), with some bytes changed, and verify that the connection is silently dropped.

Given that there are enough traffic to the server, shadowsocks-libev and shadowsocks-rust eventually passes this test with some noticeable probability. Compare this to TLS (not 0-RTT), which will never pass the first test since TLS always accepts replay of the ClientHello message. Even for TLS 0-RTT, the server returns HTTP 425 when it detects a replay, instead of silently dropping the connection, and therefore also fails to pass the first test. Therefore, a firewall can perform this test randomly, and block the server whenever all three tests passes. (The distribution of the delay can be chosen so that the amortized space complexity to carry out the attack is low.)

In fact, if the distinguisher/censor keeps track of a counter for all the new connections made to the server, he can in principle know exactly when to replay the packet that it will pass the verification, since funnily enough, the number of maximum entries for the Bloom filter is hardcoded in shadowsocks-libev/shadowsocks-rust: https://github.com/shadowsocks/shadowsocks-rust/blob/6583640d5fad717c365849797dde2502c6fd7ea7/crates/shadowsocks/src/context.rs#L10-L13

Given the data from https://github.com/net4people/bbs/issues/22, it is likely that this attack is already deployed. There are a few alternative ways to protect against this attack, but this PR is the only way without changing the protocol, and I am testing its effectiveness on a popular VPS. By removing the Bloom filter, shadowsocks-rust now fails the first test and passes the other two, which is consistent with the behavior of TLS. This backs up my claim that "doing nothing is better than having a Bloom filter," as per https://github.com/shadowsocks/shadowsocks-rust/pull/556#issuecomment-864369583.

zonyitoo:
Well, sounds reasonable. What do you think @madeye ?

Mygod:
This also fixes https://github.com/shadowsocks/shadowsocks-org/issues/184.

> Since you have consistently given downvotes/objections to proposals of any change (shadowsocks/shadowsocks-org#177, shadowsocks/shadowsocks-org#178, shadowsocks/shadowsocks-org#183 (comments)), I'm giving your PR a 👎.

@database64128 You mean I have consistently given downvotes/objections to proposals of any **useless or even harmful** change*. You are welcome.

*Also I have supported the change of adding AAD to the protocol, as per https://github.com/shadowsocks/shadowsocks-org/issues/183#issuecomment-788569701. However, since it is a breaking change, it is not worth the upgrade for now.

Also since you mentioned it, you can see that I have already proposed some similar ideas in https://github.com/shadowsocks/shadowsocks-org/issues/183#issuecomment-790285591. I think this Bloom filter thing is a good example of how various proposals of "protecting against replay attacks" or "forward secrecy" or [insert your other unnecessary demands on Shadowsocks and drafty protocol "upgrades"] can actually harm the main goal of Shadowsocks being an access tool.

database64128:
> This by itself is a feature that an active probing adversary can utilize to distinguish shadowsocks-rust from TLS by doing replay.

> By removing the Bloom filter, shadowsocks-rust now fails the first test and passes the other two, which is consistent with the behavior of TLS.

This makes no sense at all. Even a simple DPI system can tell that Shadowsocks AEAD traffic is not TLS. Not mimicking TLS doesn't make Shadowsocks unique. But a half-baked "fake" TLS will almost certainly stand out.

madeye:
@zonyitoo instead of removing the bloomfilter, I think we can disable its behavior "drop replay connection" by default.

Let's still report the replay in the log, which should help to understand the replay probe in the further.

zonyitoo:
Well, that should be a good idea.

ss-go2 also have a bf for detecting replay connections, what do you think? @riobard

zonyitoo:
https://github.com/Jigsaw-Code/outline-ss-server/blob/master/service/PROBES.md

This is how outline-ss-server handles replay connections.

rprx:
需要留意，一旦移除了重放过滤器，https://github.com/shadowsocks/shadowsocks-org/issues/183 会有更多玩法。

Mygod:
@zonyitoo There is no backwards compatibility to worry about here. The honest users should not be affected by this change at all.

@rprx Any concrete efficient attack for detecting Shadowsocks traffic if we remove the filter?

rprx:
@Mygod 

> You mean I have consistently given downvotes/objections to proposals of any useless or even harmful change*. You are welcome.

https://github.com/shadowsocks/shadowsocks-org/issues/178#issuecomment-864502126

> Any concrete efficient attack for detecting Shadowsocks traffic if we remove the filter?

比如 https://github.com/shadowsocks/shadowsocks-org/issues/183 提到的对客户端的重放，以及将客户端的请求塞回给客户端、将服务端的响应塞回给服务端、再加上一些逐字节探测的手段，还有 UDP DNS 杂交到 TCP 对比响应长度什么的...很难排列组合完，这些都变成了加密层无法识别、无法防御的攻击，具体的 Shadowsocks 受攻击响应行为 pattern 需逐个实测。

dev4u:
> I think it can also be a configuration feature, simply:
> 
> ```jsonc
> {
>     // A switch for enabling silent drop of replay connections
>     // It is "true" by default, for backward compatiblity.
>     "silent_drop_replay": false
> }
> ```

能不能判断运行在plugin模式下自动切换？

zonyitoo:
> > I think it can also be a configuration feature, simply:
> > ```js
> > {
> >     // A switch for enabling silent drop of replay connections
> >     // It is "true" by default, for backward compatiblity.
> >     "silent_drop_replay": false
> > }
> > ```
> 
> 能不能判断运行在plugin模式下自动切换？

Ah. That’s a good point.

riobard:
@zonyitoo What exactly is "silent_drop_replay" in this case? Currently go-ss2 just keeps reading from the connection after detecting replay without actively dropping it. Does ss-rust behave the same?

zonyitoo:
> @zonyitoo What exactly is "silent_drop_replay" in this case? Currently go-ss2 just keeps reading from the connection after detecting replay without actively dropping it. Does ss-rust behave the same?

Yes, they work the same. Here Mygod proposed a change to remove it completely, and madeye suggested to change the behavior to detect only. What do you think?

riobard:
@zonyitoo It's definitely better to keep the replay detection at least to inform server admins that they're being probed. 

I'm undecided as to whether or not disable silent drop. If the end goal is to make it more troublesome to probe, maybe a chaotic approach would be better? e.g. after detecting replay (or any invalid attempts to connect), we could respond randomly, i.e. sometimes drop the connection after a random timeout, sometimes keep the connection but reply with bogus data, sometimes just keep draining as if nothing is wrong. Hopefully it would confuse probes.

zonyitoo:
If respond randomly, then the attacker would try with one data stream multiple times, and check if the server respond randomly.

If we can persist the IV filter for a long time, then the current issue should be able to avoid.

Mygod:
> If we can persist the IV filter for longer time, then the current issue should be able to avoid.

Do you want to use a database holding all the past IV, and make your server super slow, and have users complaining that their disk is full?

zonyitoo:
Storing the past bloom filters should be enough. And we don't need to store all filters.

Mygod:
No and the reason is left as exercise.

Mygod:
Updated the patch to keep the error logging as per https://github.com/shadowsocks/shadowsocks-rust/pull/556#issuecomment-864476792.

P.S. The answer to the exercise above is that your running time still scales linearly, and even worse your false positive rate goes up also linearly.

riobard:
@zonyitoo 

> If respond randomly, then the attacker would try with one data stream multiple times, and check if the server respond randomly.

This is easy to counter-react: choose one of many responses based on a hash of the IV then we have the same response for one particular stream.

zonyitoo:
Nah. Since the hash algorithm is public, they can easily verify that some specific streams have exactly expected behaviors.

BTW, I cannot make a decision about this PR, because I couldn’t evaluate the influence. But since the Org have just confirmed that “shadow socks is a tunnel instead of a security layer”, this PR is acceptable.

So I want to make a poll: Do you vote for this PR? Would you follow this changes in your projects? @madeye @riobard

Mygod:
As I mentioned repeatedly, I am currently evaluating the impact of this change. I have set up a popular server running shadowsocks-libev and shadowsocks-rust. I have repeated this experiment three times (on same server, three different ports) and they all got blocked within 3 days. I am now running shadowsocks-rust without the replay filter and it has been 3 days and still going.

I am planning to give this a few more days and if all goes well, we should drop this behavior across all implementations, until someone discovers some other deployable attack.

database64128:
I vote against this change.

Mygod:
Nobody asked for your opinion, you politician.

database64128:
@Mygod What version of ss-rust did you test with the replay filter? The drain behavior had been broken all along and was only fixed in v1.11.1. Before v1.11.1, connections that were determined as replay would be closed almost immediately.

zonyitoo:
What you mentioned were these lines: https://github.com/shadowsocks/shadowsocks-rust/blob/3743ffe1c33f1f854080638262a36ce71dcfd67a/crates/shadowsocks-service/src/server/tcprelay.rs#L123-L127

It should be fine if these lines are included.

Mygod:
Now we are talking. That is a reasonable assumption but I have evidence to suggest that the attack that I proposed is more likely the one that is being used by GFW:

1. In my tests, the port only got blocked much later than when it first received replay. If GFW is trying to identify the replay behavior, the port should be blocked as soon as it received replay. The fact that there is a delay suggests that it is more likely that GFW is detecting for a ping-pong filter.
2. Although it is possible I was running an older version of shadowsocks-rust, shadowsocks-libev also got blocked where this behavior was not present.

I am keeping my plan of running the current test for a few more days. Feel free to set up a server yourself and test your assumption.

database64128:
> Nobody asked for your opinion, you politician.

That doesn't mean I can't post my opinions. If you don't want to see me here, feel free to use your "executive power" to ban me from this organization and delete all my posts. Oh, and don't forget to force push to [shadowsocks/shadowsocks-windows](https://github.com/shadowsocks/shadowsocks-windows) to get it back to https://github.com/shadowsocks/shadowsocks-windows/commit/18486788997bd4c7acc5db6b7823f75a9e5b78a5. After all, @madeye had already played the _force push and quietly kick you out_ game once.

riobard:
@zonyitoo I'm inclined to wait for more evidence. I'm running go-ss2 with anti-replay on personal servers and not getting blocked. YMMV.

Mygod:
@riobard As pointed out in the findings of https://github.com/net4people/bbs/issues/22, while all Shadowsocks receives replay probing, only popular VPS gets blocked. This is also consistent with my finding where I only reproduced this on a popular VPS. (I have also mentioned this twice before.) This makes sense as it minimizes the collateral damage while helps GFW collect data on evaluating future collateral damage if they wish to roll out this block to all servers. If you are running it on your personal server, it is likely that GFW does not block it currently and you could try it on more popular VPS to reproduce this.

madeye:
I'm fine with this PR.

But as I mentioned before many times, you need setup plugins to obfuscate your traffic, otherwise any suspicious server port without a common protocol character could be blocked in the "sensitive" timeframe (like this month).

riobard:
@Mygod Unfortunately I do not have a popular server to test. Would you mind testing go-ss2 on the same server with ss-rust/ss-libev and see if blocking can be reliably triggered? You can also disable replay protection by specifying the environment variable `SHADOWSOCKS_SF_CAPACITY=0`.

@zonyitoo Since go-ss2 has the option to disable replay protection, right now I do not need to make any code changes: users can experiment with the feature (default is on tho) to their liking.

linuxbunter213:
I dropped a shadow socks packet and rust it was there for 2 days before
there was devastation be careful and fuck that Pc cunt too .. just do it
rio

zonyitoo:
Could anyone tell me what is this @linuxbunter213 ? It seems like a robot.

gfw-report:
https://github.com/shadowsocks/shadowsocks-rust/pull/556#issuecomment-864432692:

> 1. Immediate replay of the packet, and verify that the connection is silently dropped.
> 2. Replay of the packet after a random amount of time, and verify that the server eventually accepts and returns some data.
> 3. Replay of the packet (also after a random amount of time), with some bytes changed, and verify that the connection is silently dropped.

Fascinating hypothesis, @Mygod.

https://github.com/shadowsocks/shadowsocks-rust/pull/556#issuecomment-866203809

> I am now running shadowsocks-rust without the replay filter and it has been 3 days and still going.

We have some additional observations that Shadowsocks implementations **without** the replay filter are surprisingly not blocked by the GFW for the time being.

Particularly, we have been running two servers, located outside of China, for more than a month without getting blocked. For each server, we run 7 instances of ShadowsocksR on different ports. All instances use `plain` as the obfuscation and each instance uses a different encryption method. This is equivalent to a Shadowsocks server without replay filter as ShadowsocksR never has a replay filter. We make one connection to the each listening ports using clients in China every 10 seconds.

---

Although the hypothesis is plausible and many evidences suggest that the replay filter has something to do with the blocking, we may still want to be extra cautious about removing the replay filter because:

1. We still haven't fully validated/falsified this hypothesis. In other words, we don't understand 1) what probes are sent by the censor, and 2) how the censor uses these probing information. For example, our understanding is that you attributed the fact that less popular servers had not been blocked by the GFW to their bloom filters not getting hit enough times to get reset. So what if we let a less popular server run for a day, and then manually empty the bloom filter by restarting the server? If the server are not blocked as expected, we may falsify this hypothesis (or at least know this hypothesis needs some changes).
2. We may still not fully understand the implications of removing replay filter or ["disabling its behavior 'drop replay connection'"](https://github.com/shadowsocks/shadowsocks-rust/pull/556#issuecomment-864476792) as proposed by @madeye. For example, it seems that **more distinguishable fingerprints may be exhibited under the same attack if we removed the replay filter**:

For each legitimate connection the censor observes, the censor immediately sends two probes to the server: one exact replay (*R0*); and one replay with the first byte changed (*R1*). If the server always responded to the *R0* with some random data, but never responded to *R1* with any data (because it fails the AEAD authentication), then it matches the fingerprints of a Shadowsocks server using AEAD cipher with replay filter disabled.

For reason 1, we will design and conduct experiments to validate/falsify the blocking hypothesis.

For reason 2, we may consider provide enabling/disable replay filter as a feature. And since we haven't observe any Shadowsocks server without replay filter gets blocked, we may advise users to disable it for the time being.

What do you think?

Mygod:
For reason 1, I think there is a misunderstanding. By less popular, I mean that the VPS provider itself is not popular, not that the Shadowsocks port is not getting enough connections/traffic for the active probing to work. See also: https://github.com/shadowsocks/shadowsocks-rust/pull/556#issuecomment-866274187

For reason 2, that is a valid fingerprint, but this distinguisher could lead to much larger collateral damage, compared to the current one which is essentially detecting for the presence of a **ping-pong** Bloom filter. Let me emphasize that a **ping-pong** filter is not full replay protection, and the attack I proposed essentially leverages this fact. Indeed, the attack would only filter out protocols that do use a **ping-pong** filter, which is a very distinctive feature of some Shadowsocks implementations.

To conclude, this attack shows that having a **ping-pong** filter is worse than not having replay protection at all. Your point regarding importance of having replay filters assumes that you have a **perfect** replay filter, which a ping-pong filter is far from.

Mygod:
Going back to the experiment, it is day 10 now and the new port of the same server not having ping-pong filter is still going strong.

I propose that we should merge this patch, and include this behavior as default into all implementations asap to stop servers from getting blocked. In the future, we should evaluate the possible harm of changes to the protocol/the implementations more carefully, even for something as small as a ping-pong filter.

zonyitoo:
I would still hesitate to add a compile feature as a switch. Simple as:

```rust
if (cfg!(feature = "reject-repeated-connection")) {
    return Err
} else {
    warn!(...)
}
```

but you are right, Ping-Pong bloom filter is far from a perfect mechanism for detecting repeated connections.

gfw-report:
> For reason 1, I think there is a misunderstanding. By less popular, I mean that the VPS provider itself is not popular, not that the Shadowsocks port is not getting enough connections/traffic for the active probing to work. See also: #556 (comment)

Got you. Would you mind sharing the name and the location of the VPSes you used publicly or [privately](https://gfw.report/gfw_report.asc) to us? This will greatly help us to reproduce the blocking.

Mygod:
> I would still hesitate to add a compile feature as a switch. Simple as:
> 
> ```rust
> if (cfg!(feature = "reject-repeated-connection")) {
>     return Err
> } else {
>     warn!(...)
> }
> ```
> 
> but you are right, Ping-Pong bloom filter is far from a perfect mechanism for detecting repeated connections.

Let me get this straight. Ping-pong filters are so bad, there is no rational reason to keep them, even as an option. If you are still hesitant, please just read everything I said in this PR all over again.

zonyitoo:
Would be released in v1.11.2 after the PR in tokio is merged.

riobard:
@gfw-report I think we need to figure out if long-term replay protection is in general good or bad for fingerprinting Shadowsocks servers. If you have spare resources, may I suggest that you do an analysis of existing popular non-TLS encrypted traffic and see what's the percentage of them being replay-resistant? Ideally we'd like to blend in with those traffic and respond to replay in similar ways.

I also agree with @Mygod that the existing ping-pong filter can only resist against short-term replay. If after a survey of existing traffic we do need to have proper replay protection, there're still simple fixes we can apply. Without handshake, long-term replay protection needs coarse-synced clock between clients and servers. I guess majority of our users are probably using NTP clock now so it's relatively safe to assume correct time on both sides. In another issue thread (I couldn't find now) I proposed to slightly adjust how we generate salt/IV to embed such timestamp without changing wire format of the protocol. It's also cheap and easy for the client to defend against replay after simple changes.

database64128:
For those who prefer not to have this PR merged, a fork has been created at [Shadowsocks-NET/shadowsocks-rs](https://github.com/Shadowsocks-NET/shadowsocks-rs). The fork will follow all upstream updates other than the ones that blindly disable replay protection without substitutes.

gfw-report:
> I think we need to figure out if long-term replay protection is in general good or bad for fingerprinting Shadowsocks servers. If you have spare resources, may I suggest that you do an analysis of existing popular non-TLS encrypted traffic and see what's the percentage of them being replay-resistant? Ideally we'd like to blend in with those traffic and respond to replay in similar ways.

Great suggestions, @riobard ! We have been taking actions since you suggested.

database64128:
I want to apologize for my previous statements against Mygod on replay protection. After reevaluating the situation with @zonyitoo, I now agree that the current implementation is inherently broken, and should probably be completely removed. I have retracted my downvotes and changed some of my previous statements to be hidden.

riobard:
What's the latest development?

database64128:
> What's the latest development?

It has been completely removed in 90f05a5.

riobard:
I saw it. I'm wondering what changed your mind.

zonyitoo:
1. `sslocal` in this project supports multiple remote servers. It will "ping" all the remote servers every 10 seconds. So if you have many servers, then the bloom filters will soon become full and start to cleanup themselves.
2. Many websites and Apps uses QUIC. Every UDP packets will need to generate an unique salt/iv, that will also cause the bloom filters become full in a very short time (a couple of minutes).
3. If a `ssserver` is shared by multiple clients, then each individual clients may generate the same salt/iv then be rejected by the `ssserver`.
4. The SYN of TFO may be retransmitted, then the `ssserver` will see this as two individual clients and reject the second one (the real one).

database64128:
> I saw it. I'm wondering what changed your mind.

We first observed a high false positive rate on TFO-enabled servers. This was probably partially caused by [duplicate SYNs with data](https://datatracker.ietf.org/doc/html/rfc7413#section-6.1). It has led us to do a little math on the real world effectiveness of the replay protection implementation.

Both shadowsocks-rust and Outline use 10000 as the IV cache capacity by default. Turns out you can easily exhaust it in as little as a few minutes even on your personal server. Just watch a YouTube video via QUIC (It's the default transport on both desktop and mobile). Half way through the video, the cached IVs have probably already been rotated.

Even if you give up on filtering UDP traffic (like Outline does), new TCP connections from normal household Internet usage can exhaust your IV cache capacity in less than a few hours. Let's also not forget about the fact that most of us share servers with family, friends or in a community.

Clients also can't guarantee the uniqueness of the generated salt due to the limited capacity, leading to higher false positive rate at server side. With 250 servers in ss-rust's ping balancer, it takes exactly 5 minutes to fill up the salt cache.

riobard:
UDP is an afterthought anyway, so it's fine to ignore those. 10k capacity is clearly insufficient. IIRC libev port defaults to 1m? You should raise the capacity on busy servers.

I'm wondering if there's any behavior leaks. Previous discussion seems resulted in no clear conclusion.

riobard:
> Also, the bloom filter can only allow one client to one server configuration. But in reality, users have mobile phones, PCs, smart routers, relays, ... The current implementation will require all clients to generate salt/iv to be globally unique, which is impossible in most cases.

What are you even thinking? Which part of UUID have you forgotten about? :D

zonyitoo:
1. Raising up capacity can help but not an ultimate solution.
2. Ok, then we can disable UDP replay protection, but QUIC is now become more and more popular in the Internet. So in the end, we have already disable replay detect for the most cases already.
3. It is not a capacity issue. Different clients won't be able to generate globally unique random iv/salts without communicating. If you use UUID as iv/salt, then it would become a completely fault.
4. Maybe not, TFO is not broken. You can get obvious latency improvement after enabling TFO.

riobard:
I'm sorry but this is not the constructive way to reason about it. 

Probabilistic approaches like Bloom filters will _never_ give you 100% reliable results. The idea is to find a very low probability such that it won't matter in practice. If all you're concerned about is false positive rate, just increase the capacity at  the cost of cheap RAM will solve the problem. There's no ultimate solutions. Just tradeoffs.

UDP traffic has completely different behaviors. Remember we need replay protection for fingerprinting. Now convince me if QUIC can be fingerprinted by replaying? If not, it's fine to leave it alone.

Once again, tell me what's the probability of two honest clients repeatedly generating 256-bit randomness and ending up colliding? It's not a capacity issue coz it's a statistics issue. Use your math! 😂

Even if TFO is not broken, how do you deal with the fact that almost no other TCP traffic adopts TFO? What replay protection protects us from? It's completely backwards.

Again, I'm not saying replay protection is alright. I'm just not seeing the correct arguments _against_ it.

zonyitoo:
Or we should keep it simple:

1. The Ping-pong Bloom filter is not a perfect solution because it have obvious flaws, and you have agreed with it.
2. It couldn't protect us from replay attack because it cannot distinguish legitimate clients and evil clients.
3. The experiment that Mygod have done is that the current solution have no effect in protecting servers from being discovered by GFW.

Just that simple. Ineffective, False Positive, Flawed.

We can further engineered a new solution against replay attack.

riobard:
Bloom filters obviously is not perfect, but a) there's no perfect solution; and b) SS itself is not a perfect solution; and c) I'm not saying we should not use an imperfect solution. C'est la vie! 

>  it cannot distinguish legitimate clients and evil clients.

You're jumping right into the conclusion without even arguing for it. How does it not distinguish replayed traffic? Sure it cannot protect us from every attack but for the intended protection against short-term replay it's very effective. 

@Mygod's experiment above is very valuable but it's only a single data point. So far there's no more useful data points to (dis)prove the effectiveness of replay protection at all. @gfw-report agreed to do more experiments but it might take more time to reach conclusions.

My objection to your arguments so far is that they're mostly nitpicking about configuration defects (which can be simply corrected by changing some parameters), and some of they are factually wrong (e.g. 256-bit randomness collision) or conflicting objectives (e.g. TFO vs hidden-in-plain-sight).

Mygod:
I don't think 90f05a5a0bd655e550b1ba571092ada0ed201c07 is necessary. As proposed by @madeye, detecting the replays and printing a warning (without acting on the packet differently in a way observable by the external environment) is still useful.

@riobard Certainly 10000 capacity is not enough. However, considering that all it takes for an attacker to distinguish is a single replay failure, you would have to remember virtually every single past IV to protect against this attack. On the other hand, the attack only needs to randomly pick a constant number of IVs to remember and eventually succeed at detecting. Removing the bloom filter invalidates this attack completely without the server remembering a single IV.

As I have stressed multiple times in this thread, Bloom filters are so shit that it is better not having it to begin with. Good bye.

zonyitoo:
I am not trying to convince you because I agree with your opinion.

You are saying that this is rather an effective solution, but we were saying that we would better remove it completely than have a solution that have several obvious flaws. You may notice that we are talking about the same thing from different view.

Mygod is right, if you are actually want to protect against from replay attack, you will have to remember all IVs instead of a "short-term replay". If it can only protect in a short-term, in my point of view, it is ineffective.

riobard:
Now we're on the right path 😄

So at the core of the issue is about whether replay protection is desirable at all. The answer to that question really depends on analysis of other popular anonymous traffic and see what kind of replay protection (if any) they employ. I don't have a clear idea and I'm hoping @gfw-report could continue researching on it.

If it's deemed necessary to have _perfect_ replay protection, I agree with both of you that the current solution is very problematic. Specifically, @Mygod said that
>  the attack only needs to randomly pick a constant number of IVs to remember and eventually succeed at detecting.
which I'm not 100% sure. Yes it's definitely a weakness, but do other less-than-perfectly-replay-protected protocols exhibit similar weakness? In other words, how reliable can a probe fingerprint SS servers knowing this behavior quirk, and at what storage/scalability cost? Also
> Removing the bloom filter invalidates this attack completely without the server remembering a single IV.
which I completely agree, but then we're vulnerable to short-term replay. Are we now more vulnerable to fingerprinting or not? I guess it also depends on whether other popular anonymous traffic are susceptible to replay, right? 

Once again, I'm not pro or against replay protection. I'm just saying that we don't know enough to have a clear-cut conclusion yet.

riobard:
By the way, it's relatively easy to fix the missing long-term replay protection while keeping the simplistic Bloom filters _and_ without changing the underlying protocol. Just mix-in a coarse-grained timestamp (e.g. 15 minutes) into the salt randomization process. It would require both clients and servers to adapt for proper enforcement, but servers could choose to fallback to non-timestamped as before for backwards compatibility. 

I'm hoping we're not throwing the baby out with bathwater. 😂

zonyitoo:
You are right, I am also thinking about this kind of solution.

riobard:
Great! And if there's new solution to address it, we should probably first discuss what compatibility cost we're gonna swallow 🤢

dev4u:
2FA，可以考虑用来产生salt

ditsing:
I know one of the requirements of reply attack mitigation is to not modify the original protocol. But if we swap the salt used by the client and the server, we could trivially make reply attacks impossible. We could do the follows.

The client initiates a connection and sends a random salt to the server. The server responses with another random salt to the client. After the encryption handshake, the server uses the salt designated by the client's first request to encrypt traffic sent to the client, and the client should use the salt provided by the server to encrypt traffic sent to the server. The wire format will stay the same. Reply attack would be impossible, because the server will send different salt to different clients.

zonyitoo:
But no forward security.

BTW, that will require the server to respond instantly after received the first packet. The current shadowsocks protocol doesn't require that behavior.

ditsing:
> But no forward security.

I spent quite some time Googling this term (and [forward secrecy](https://en.wikipedia.org/wiki/Forward_secrecy)), but could not quite understand it in the context of Shadowsocks. Do you have an example scenario? 

> BTW, that will require the server to respond instantly after received the first packet. The current shadowsocks protocol doesn't require that behavior.

That is true. Not replying immediately makes the shadow server look more like an HTTP server.

PR #567, opened 2021-07-06
Remove the reuse for UNIX stream DNS lookup.

madeye:
Our UNIX stream based DNS lookup doesn't support connection reuse.

### DIFF
--- a/crates/shadowsocks-service/src/local/dns/client_cache.rs
+++ b/crates/shadowsocks-service/src/local/dns/client_cache.rs
@@ -1,7 +1,7 @@
 //! DNS Client cache
 
 #[cfg(unix)]
-use std::path::{Path, PathBuf};
+use std::path::Path;
 use std::{
     collections::{hash_map::Entry, HashMap, VecDeque},
     future::Future,
@@ -23,8 +23,6 @@ use super::upstream::DnsClient;
 enum DnsClientKey {
     TcpLocal(SocketAddr),
     UdpLocal(SocketAddr),
-    #[cfg(unix)]
-    UnixStream(PathBuf),
     TcpRemote(Address),
     UdpRemote(Address),
 }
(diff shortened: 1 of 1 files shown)


Mygod:
My compiler complains that this commit cannot be found: https://github.com/zonyitoo/tokio/commit/47b76a5a

madeye:
I got the compilation error as well, you may have to rebase to an old commit.

zonyitoo:
> My compiler complains that this commit cannot be found: [zonyitoo/tokio@47b76a5](https://github.com/zonyitoo/tokio/commit/47b76a5a)

Just ignore it. I will fix it until tokio's PR was merged.

Mygod (approved):
Can confirm this fixes everything. (I thought I was going to implement a DNS cache in shadowsocks-android for a moment so thank god!)

zonyitoo:
The UNIX stream should report errors if the endpoint have been closed. Hmm.. Why it is the cause of shadowsocks/shadowsocks-android#2751 ?

zonyitoo (approved):
LGTM. But why?

madeye:
Our android UNIX stream resolve doesn't support connection reuse yet.

I will use this PR to create a new release of shadowsocks-android as a hotfix. But for now, I think it's okay to keep it open for shadowsocks-rust.

Mygod:
There is no reason to support connection reuse on local sockets since the delay is negligible. Currently, shadowsocks-android should close the socket (syscall `close`) after the response is written.

zonyitoo:
Well, you are right about the cost of creating unix local stream is negligible. I am Ok with this PR, but I still cannot understand why it cause that issue.

Currently the implementation will open a new client if the connection have been lost due to `for _ in 0..self.retry_count`.

Mygod:
Maybe we merge this and figure out the retry later? 😛

zonyitoo:
I am worrying about if there are any issues in this connection management strategy, then it will also affects TCP and UDP DNS clients.

zonyitoo:
Will shadowsocks-android implements connection reuse in the future?

Mygod:
> There is no reason to support connection reuse on local sockets since the delay is negligible. Currently, shadowsocks-android should close the socket (syscall `close`) after the response is written.
> 
> EDIT: Holding a local socket open also wastes system resources methinks.

^

PR #574, opened 2021-07-13
Add Docker and its GitHub Actions auto build support

### LINKED ISSUE #570: provide shadowsocks-rust docker image?

kallydev:
Resolved #570. Provide multi-arch Docker support for shaodowsocks-rust.

## TODO

- [x] Determine how to provide `ssserver` and `sslocal`, etc.
- [x] Improve CI build time.

### DIFF
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -1,0 +1,36 @@
+name: Build Docker Images
+on:
+  push:
+    tags:
+      - v*
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        bin: 
+          - ssserver
+          - sslocal
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Setup Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v1 
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Docker metadata
+        id: metadata
+        uses: docker/metadata-action@v3
+        with:
+          images: ghcr.io/${{ github.repository_owner }}/${{ matrix.bin }}-rust
+      - name: Build and release Docker images
+        uses: docker/build-push-action@v2
+        with:
+          platforms: linux/386,linux/amd64,linux/arm64/v8
+          target: ${{ matrix.bin }}
+          tags: ${{ steps.metadata.outputs.tags }}
+          push: true
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,50 @@
+FROM --platform=$BUILDPLATFORM rust:1.53.0-buster AS build
+
+ARG TARGETARCH
+
+RUN apt-get update && apt-get install -y build-essential curl musl-tools
+
+WORKDIR /root/shadowsocks-rust
+
+ADD . .
+
+RUN rustup install "$(cat rust-toolchain)" && \
+    case "$TARGETARCH" in \
+    "386") \
+        RUST_TARGET="i686-unknown-linux-musl" \
+        MUSL="i686-linux-musl" \
+        ;; \
+    "amd64") \
+        RUST_TARGET="x86_64-unknown-linux-musl" \
+        MUSL="x86_64-linux-musl" \
+        ;; \
+    "arm64") \
+        RUST_TARGET="aarch64-unknown-linux-musl" \
+        MUSL="aarch64-linux-musl" \
+        ;; \
+    *) \
+        echo "Doesn't support $TARGETARCH architecture" \
+        exit 1 \
+        ;; \
+    esac && \
+    wget -qO- "https://musl.cc/$MUSL-cross.tgz" | tar -xzC /root/ && \
+    CC=/root/$MUSL-cross/bin/$MUSL-gcc && \
+    rustup target add $RUST_TARGET && \
+    RUSTFLAGS="-C linker=$CC" CC=$CC cargo build --target $RUST_TARGET --release && \
+    mv target/$RUST_TARGET/release/ss* target/release/
+
+FROM alpine:3.14 AS sslocal
+
+COPY --from=build /root/shadowsocks-rust/target/release/sslocal /usr/bin
+
+COPY --from=build /root/shadowsocks-rust/examples/config.json /etc/shadowsocks-rust/
+
+ENTRYPOINT [ "sslocal", "--log-without-time", "-c", "/etc/shadowsocks-rust/config.json" ]
+
+FROM alpine:3.14 AS ssserver
+
+COPY --from=build /root/shadowsocks-rust/target/release/ssserver /usr/bin
+
+COPY --from=build /root/shadowsocks-rust/examples/config.json /etc/shadowsocks-rust/
+
+ENTRYPOINT [ "ssserver", "--log-without-time", "-c", "/etc/shadowsocks-rust/config.json" ]
--- a/examples/config.json
+++ b/examples/config.json
@@ -5,5 +5,5 @@
     "local_address": "127.0.0.1",
     "password": "password",
     "timeout": 300,
-    "method": "aes-256-cfb"
+    "method": "aes-256-gcm"
 }


kallydev:
1. shadowsocks-rust has multiple binary executables, and we should also provide `sslocal` and `ssserver` separately, just as frp provides [frpc](https://hub.docker.com/r/fatedier/frpc/tags) and [frps](https://hub.docker.com/r/fatedier/frps/tags) images.

```bash
docker run shadowsocks/sslocal-rust:latest
docker run shadowsocks/ssserver-rust:latest
```

2. Does shadowsocks-rust have a Docker Hub account token for the shadowsocks organization? Or we can just publish Docker images on the GitHub Container Registry.

3. Since rustup uses `gnu` instead of `musl` by default, and they run almost indistinguishably in Docker, we don't need to provide a `musl` version.

Please help me check if I am going in the right direction, thanks. @zonyitoo

zonyitoo:
Actually we have `sslocal`, `ssserver` and `ssmanager`. It is Ok to have multiple images for locals and servers.

I don't think shadowsocks organization has a special Docker Hub account. Do we? @madeye 

It should be Ok to publish to Github's Docker Hub. (Is it free for open source software?)

> we don't need to provide a musl version.

Yes, it is.

BTW, how to use it in different architectures? For example, how can I use it on ARM based computers?

madeye:
https://hub.docker.com/orgs/shadowsocks I can add you to this group.

madeye:
BTW, dockerhub limits 3 members per organization for any free account. I think you may try Github's service instead.

kallydev:
> Is it free for open source software?

Yes, [GitHub Packages is free for public repositories](https://github.com/features/packages#pricing).

> BTW, how to use it in different architectures? For example, how can I use it on ARM based computers?

Docker Hub automatically provides mirrors for the corresponding architecture. But GitHub Container Registry does not provide the same feature, you must specify the `SHA-256` value after the `tag` when pulling the image. I have an arm64 device and I will make sure that it works.

The specific content can be seen in the [official demo picture](https://github.blog/changelog/2020-09-17-github-container-registry-support-for-multi-arch-images/).

```bash
# Docker Hub
# Linux/amd64 and Linux/arm64
docker pull shadowsocks-rust/ssserver-rust

# GitHub Container Registry
# Linux/amd64
docker pull ghcr.io/shadowsocks-rust/ssserver-rust:latest@sha256:76f015d59c2bed9bc4919e22031029e96c9ed287ae2e025365c8277e46cfd1b5
# Linux/arm64
docker pull ghcr.io/shadowsocks-rust/ssserver-rust:latest@sha256:8495f7941b345e333266d3ed9b99ad976b52dff311b61e08e826c84f7d149963
```

> I think you may try Github's service instead.

OK, let's try GitHub Container Registry first.

Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+
+RUN apk add build-base curl
+
+RUN curl https://sh.rustup.rs -o rustup-init.sh && sh rustup-init.sh -y --default-toolchain nightly
+
+WORKDIR /root/shadowsocks-rust
+
+ADD . .
+
+RUN cargo build --release --bin sslocal
```
database64128, line 13: Just run `cargo build --release`. It starts building all binaries at the same time when all dependencies are built, so it can take advantage of multiple CPU threads. And it also gives you `ssurl` and `ssmanager`.

kallydev: Yes, your solution can be effective in reducing time when building all binaries. But this Dockerfile is only for users building on their own, and they often only need one platform and one binary.

My plan is to have the CI use a separate Dockerfile, because performance is extremely poor when the CI uses Docker BuildX with QEMU to build ARM images. For example, building an amd64 image takes only 7 to 10 minutes, while building an arm64 image takes over 40 minutes, which is unacceptable.

I will later provide [cross](https://github.com/rust-embedded/cross) based builds for CI to solve this problem. After that I will implement your solution, compare them and choose the best one, thank you for your advice.

.github/workflows/build-docker-image.yml
```diff
@@ -0,0 +1,38 @@
...
+        uses: docker/login-action@v1 
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Docker metadata
+        id: metadata
+        uses: docker/metadata-action@v3
+        with:
+          images: ghcr.io/${{ github.repository_owner }}/${{ matrix.bin }}-rust
```
zonyitoo, line 31: Is the postfix `-rust` really necessary? Could we tag the image with name like `ghcr.io/shadowsocks-rust/sslocal`?

kallydev: Since the full name format of the GCR image is `ghcr.io/OWNER/IMAGE_NAME:latest`, images belong to OWNER not repository, they are just bound to this repository, so the OWNER can only be `shadowsocks`. I added the rust suffix to prevent conflicts with the naming of [shadowsocks-libev](https://github.com/shadowsocks/shadowsocks-libev)'s sserver and sslocal.

This is a limitation of the GCR and I have no way to modify it.

examples/config.json
```diff
@@ -5,5 +5,5 @@
     "local_address": "127.0.0.1",
     "password": "password",
     "timeout": 300,
```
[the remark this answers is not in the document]
kallydev: I'm not sure, this is just a file for testing, do we need to delete the timeout parameter?  @zonyitoo

kallydev:
It's now working on `i386`, `amd64`, `arm64`. Since I don't have an `armv7` device to use for build testing, I didn't rush to add support for it. The build time for Docker images has been optimized from 1 hour to 23 minutes, and if we don't add `i386` support, the whole build process will take only 16 minutes.

I think the optimal solution for build time is under 8 minutes, but that relies on Docker's experimental [manifest](https://docs.docker.com/engine/reference/commandline/manifest/) command, which is not officially recommended for use in production environments. However, when the `manifest` command is included in a stable version of Docker, I may be able to provide an improved solution for current image builds.

If you are building the image yourself, refer to these steps.

```bash
docker buildx build -t shadowsocks/ssserver-rust:latest -t shadowsocks/ssserver-rust:v1.11.1 --target ssserver .
docker buildx build -t shadowsocks/sslocal-rust:latest -t shadowsocks/sslocal-rust:v1.11.1 --target sslocal .
```

```bash
kallydev@ubuntu:~/Developer/shadowsocks-rust$ docker images
REPOSITORY                  TAG             IMAGE ID       CREATED          SIZE
shadowsocks/sslocal-rust    latest          b28413750c94   10 seconds ago   15.8MB
shadowsocks/sslocal-rust    v1.11.1         b28413750c94   10 seconds ago   15.8MB
shadowsocks/ssserver-rust   latest          ddfe4b494ab4   5 seconds ago    13.5MB
shadowsocks/ssserver-rust   v1.11.1         ddfe4b494ab4   5 seconds ago    13.5MB
<none>                      <none>          36c3a5c6c661   14 minutes ago   1.28GB
rust                        1.53.0-buster   61281569db97   3 weeks ago      1.25GB
```

If you want to use an image that was built with CI, please refer to these steps.

```bash
docker run --name sslocal-rust \
  --restart always \
  -p 1080:1080/tcp \
  -v /opt/shadowsocks-rust/config.json:/etc/shadowsocks-rust/config.json \
  -dit ghcr.io/kallydev/sslocal-rust

docker run --name ssserver-rust \
  --restart always \
  -p 8388:8388/tcp \
  -p 8388:8388/udp \
  -v /opt/shadowsocks-rust/config.json:/etc/shadowsocks-rust/config.json \
  -dit ghcr.io/kallydev/ssserver-rust
```

```bash
kallydev@ubuntu:~/Desktop/shadowsocks-rust$ docker ps
CONTAINER ID   IMAGE                            COMMAND                  CREATED              STATUS          PORTS                                                                                  NAMES
14c80a5392dd   ghcr.io/kallydev/sslocal-rust    "sslocal --log-witho…"   51 seconds ago       Up 49 seconds   0.0.0.0:1080->1080/tcp, :::1080->1080/tcp                                              sslocal-rust
a7af8d251b65   ghcr.io/kallydev/ssserver-rust   "ssserver --log-with…"   About a minute ago   Up 59 seconds   0.0.0.0:8388->8388/tcp, 0.0.0.0:8388->8388/udp, :::8388->8388/tcp, :::8388->8388/udp   ssserver-rust

kallydev@ubuntu:~/Desktop/shadowsocks-rust$ docker images
REPOSITORY                       TAG       IMAGE ID       CREATED         SIZE
ghcr.io/kallydev/sslocal-rust    latest    e98bbf06d042   4 minutes ago   15.8MB
ghcr.io/kallydev/ssserver-rust   latest    bdda06e583e9   8 minutes ago   13.5MB
```

I've provided a demo of GitHub Action and Packages.

https://github.com/kallydev/shadowsocks-rust-patch/actions/runs/1052603426
https://github.com/kallydev?tab=packages&repo_name=shadowsocks-rust-patch

zonyitoo:
What's the tag of the images when building with actions?

kallydev:
> What's the tag of the images when building with actions?

The image tag is the same as the git tag.

zonyitoo:
Thanks for you contribution.

zonyitoo:
Do you interested in writing a detailed document in `README.md` for instructing users about how to use the Docker services?

Issue #581, opened 2021-08-01
Support tun to remove the dependency on tun2socks.


FH0:
The following example is very simple, and does not rely on or implement tcp/ip stack.
https://github.com/FH0/stn/blob/main/stn_tun/src/tcp.rs#L48-L58
https://github.com/FH0/stn/blob/main/stn_tun/src/tcp.rs#L161-L166

zonyitoo:
Here is another possible solution: https://github.com/shadowsocks/tun2socks . If we decide to provide tun interface, then we have to consider:

- cross platform support, Linux-like, macOS, BSD-like, Windows
- DNS solution, port 53 capture? Fake DNS?
- Extended ACL rules based on geographic information
- TCP connection lifetime management
- …

Currently I may not have time to fully investigate an ultimate solution. It is welcome to make a detailed design and merge it into the next release.

FH0:
- https://github.com/shadowsocks/tun2socks is not as good as in the above
example. It depends on the crate of a tcp/ip stack.
https://github.com/shadowsocks/tun2socks/blob/master/Cargo.toml#L22
- For DNS, the solution is the same as TPROXY.
- TCP and ACL are not a problem. The above example just redirects the data
packet. But notice here:
https://github.com/FH0/stn/blob/main/stn_tun/src/tcp.rs#L133-L142.
- Maybe Linux can be supported first, and other systems will be supported
in later versions.

ty ***@***.***> 于2021年8月1日周日 下午12:46写道：

FH0:
- https://github.com/shadowsocks/tun2socks is not as good as in the above
example. It depends on the crate of a tcp/ip stack.
https://github.com/shadowsocks/tun2socks/blob/master/Cargo.toml#L22
- For DNS, the solution is the same as TPROXY.
- TCP and ACL are not a problem. The above example just redirects the data
packet. But notice here:
https://github.com/FH0/stn/blob/main/stn_tun/src/tcp.rs#L133-L142.
- Maybe Linux can be supported first, and other systems will be supported
in later versions.

zonyitoo:
It seems that your code (POC) didn't handle TCP's retransmission, packet reordering, ... Since shadowsocks runs on TCP / UDP (transmission layer), if you didn't include a fully functional TCP network stack, then it may cause data transmitting error.

FH0:
I'm sorry that I didn't give a detailed explanation to make you misunderstood.

[...]

PR #586, opened 2021-08-04
Local server support Tun interface for both TCP and UDP

zonyitoo:
ref #581

### TCP

image

1. Create a `TcpListener` listening on an address in the subnet of `tun` device.
2. When `tun` receives TCP packets, it will redirect it to `TcpListener` by modifying `saddr` to a unique fake address and `daddr` to the address that `TcpListener` was listening.
3. `TcpListener` accepts a new connection. The connection's `peer_addr` will be the unique fake address. So we can get the original destination address by looking to the `tun`'s address allocation map.

### UDP

The same as the `redir` design, except it sends and receives packets from the `tun` interface.

### Possible Applications

- shadowsocks-android
- shadowsocks-ios
- macOS & Linux
- Windows (with the [Wintun](https://www.wintun.net/) project)


dev4u:
噢，又有新玩法了！
什么时候可以尝鲜测试？

zonyitoo:
This is already available for trial.

zonyitoo:
> 按照主题的结构图，plugin会在什么位置上呢？

Just like a normal `sslocal`, runs after the `TcpListener`.

dev4u:
> This is already available for trial.

好，我找时间试试。

dev4u:
```log
error[E0061]: this function takes 1 argument but 2 arguments were supplied
   --> bin/sslocal.rs:359:48
    |
359 |                     let mut local_dns_config = LocalConfig::new(addr, ProtocolType::Dns);
    |                                                ^^^^^^^^^^^^^^^^ ----  ----------------- supplied 2 arguments
    |                                                |
    |                                                expected 1 argument
    |
note: associated function defined here
   --> /Users/anders/develop/workspaces/shadowsocks-android/core/src/main/rust/shadowsocks-rust/crates/shadowsocks-service/src/config.rs:693:12
    |
693 |     pub fn new(protocol: ProtocolType) -> LocalConfig {
    |            ^^^
```
放到android里编译失败了。

dev4u:
```log
> Task :core:cargoBuildArm64
   Compiling shadowsocks-service v1.12.0 (/workspaces/shadowsocks-android/core/src/main/rust/shadowsocks-rust/crates/shadowsocks-service)
error[E0432]: unresolved import `tun::AsyncDevice`
  --> crates/shadowsocks-service/src/local/tun/mod.rs:21:11
   |
21 | use tun::{AsyncDevice, Configuration as TunConfiguration, Device, Layer};
   |           ^^^^^^^^^^^ no `AsyncDevice` in the root

error[E0425]: cannot find function `create_as_async` in crate `tun`
  --> crates/shadowsocks-service/src/local/tun/mod.rs:89:33
   |
89 |         let device = match tun::create_as_async(&self.tun_config) {
   |                                 ^^^^^^^^^^^^^^^ not found in `tun`

error[E0599]: no method named `poll` found for struct `Pin<&mut _>` in the current scope
   --> crates/shadowsocks-service/src/local/tun/mod.rs:165:13
    |
165 | /             tokio::select! {
166 | |                 // tun device
167 | |                 n = self.device.read(&mut packet_buffer) => {
168 | |                     let n = n?;
...   |
197 | |                 }
198 | |             }
    | |_____________^ method not found in `Pin<&mut _>`
    |
    = note: `fut` is a function, perhaps you wish to call it
    = note: this error originates in the macro `$crate::select` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0599]: no method named `poll` found for struct `Pin<_>` in the current scope
   --> crates/shadowsocks-service/src/local/tun/mod.rs:165:13
    |
165 | /             tokio::select! {
166 | |                 // tun device
167 | |                 n = self.device.read(&mut packet_buffer) => {
168 | |                     let n = n?;
...   |
197 | |                 }
198 | |             }
    | |_____________^ method not found in `Pin<_>`
    |
    = note: `fut` is a function, perhaps you wish to call it
    = note: this error originates in the macro `$crate::select` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0599]: no method named `packet_information` found for mutable reference `&mut tun::platform::Configuration` in the current scope
  --> crates/shadowsocks-service/src/local/tun/mod.rs:86:24
   |
86 |             tun_config.packet_information(false);
   |                        ^^^^^^^^^^^^^^^^^^ method not found in `&mut tun::platform::Configuration`

warning: unused import: `AsyncReadExt`
unused import: `AsyncReadExt`

  --> crates/shadowsocks-service/src/local/tun/mod.rs:18:10
   |
18 |     io::{AsyncReadExt, AsyncWriteExt},
   |          ^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: unused import: `AsyncWriteExt`
unused import: `AsyncWriteExt`

  --> crates/shadowsocks-service/src/local/tun/mod.rs:18:24
   |
18 |     io::{AsyncReadExt, AsyncWriteExt},
   |                        ^^^^^^^^^^^^^

warning: unused import: `Device`
unused import: `Device`

  --> crates/shadowsocks-service/src/local/tun/mod.rs:21:59
   |
21 | use tun::{AsyncDevice, Configuration as TunConfiguration, Device, Layer};
   |                                                           ^^^^^^

Some errors have detailed explanations: E0425, E0432, E0599.
For more information about an error, try `rustc --explain E0425`.
warning: `shadowsocks-service` (lib) generated 3 warnings
`shadowsocks-service` (lib) generated 3 warnings

error: could not compile `shadowsocks-service` due to 5 previous errors; 3 warnings emitted

could not compile `shadowsocks-service` due to 5 previous errors; 3 warnings emitted

> Task :core:cargoBuildArm64 FAILED

Execution failed for task ':core:cargoBuildArm64'.
> Process 'command 'cargo'' finished with non-zero exit value 101

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

```

zonyitoo:
Did you change the `tun` dependency in `Cargo.toml`?

dev4u:
> Did you change the `tun` dependency in `Cargo.toml`?

我没有改动`Cargo.toml`，但是有修改`shadowsocks-android/core/build.gradle.kts`：
```
featureSpec.noDefaultBut(arrayOf(
        "multi-threaded",
        "stream-cipher",
        "aead-cipher-extra",
        "logging",
        "local-flow-stat",
        "local-dns",
        "local-tun",
        "jemalloc"))
```

zonyitoo:
https://github.com/meh/rust-tun/blob/19906c006512acf805f392e744e38aec8adc6618/src/lib.rs#L30-L39

Very interesting.

dev4u:
> https://github.com/meh/rust-tun/blob/19906c006512acf805f392e744e38aec8adc6618/src/lib.rs#L30-L39
> 
> Very interesting.

看来现阶段，不能运行在安卓下……

zonyitoo:
It should work on Android. Since the original crate didn't release a new version that supports tokio v1.x, I should make my `AsyncDevice` manually without its support.

ref https://github.com/shadowsocks/shadowsocks-rust/pull/586

dev4u:
> It should work on Android. Since the original crate didn't release a new version that supports tokio v1.x, I should make my `AsyncDevice` manually without its support.
> 
> ref https://github.com/meh/rust-tun/issues/29

嗯嗯，我等你的消息。谢谢！

zonyitoo:
It is now confirmed working on macOS. Would you like to test if it is working on shadowsocks-android? @madeye @Mygod

dev4u:
> It is now confirmed working on macOS. Would you like to test if it is working on shadowsocks-android? @madeye @Mygod 

现在能编译通过了吗？

dev4u:
```
   Compiling tun v0.5.1
error[E0432]: unresolved import `crate::platform::create`
  --> /.cargo/registry/src/github.com-1ecc6299db9ec823/tun-0.5.1/src/lib.rs:28:9
   |
28 | pub use crate::platform::create;
   |         ^^^^^^^^^^^^^^^^^^^^^^^ no `create` in `platform`

error[E0412]: cannot find type `Configuration` in module `platform`
  --> /.cargo/registry/src/github.com-1ecc6299db9ec823/tun-0.5.1/src/configuration.rs:38:36
   |
38 |     pub(crate) platform: platform::Configuration,
   |                                    ^^^^^^^^^^^^^ not found in `platform`
   |
help: consider importing this struct
   |
15 | use crate::Configuration;
   |

error[E0412]: cannot find type `Configuration` in module `platform`
  --> /.cargo/registry/src/github.com-1ecc6299db9ec823/tun-0.5.1/src/configuration.rs:55:34
   |
55 |         F: FnOnce(&mut platform::Configuration),
   |                                  ^^^^^^^^^^^^^ not found in `platform`
   |
help: consider importing this struct
   |
15 | use crate::Configuration;
   |

error[E0425]: cannot find value `AF_INET` in this scope
   --> /.cargo/registry/src/github.com-1ecc6299db9ec823/tun-0.5.1/src/platform/posix/sockaddr.rs:42:31
    |
42  |         if value.sa_family != AF_INET {
    |                               ^^^^^^^
    |
   ::: /.cargo/registry/src/github.com-1ecc6299db9ec823/libc-0.2.98/src/unix/linux_like/mod.rs:679:1
    |
679 | pub const AF_INET: ::c_int = 2;
    | ------------------------------- similarly named constant `_AF_INET` defined here
    |
help: a constant with a similar name exists
    |
42  |         if value.sa_family != _AF_INET {
    |                               ^^^^^^^^
help: consider importing this constant
    |
15  | use libc::AF_INET;
    |

error[E0425]: cannot find value `AF_INET` in this scope
   --> /.cargo/registry/src/github.com-1ecc6299db9ec823/tun-0.5.1/src/platform/posix/sockaddr.rs:65:27
    |
65  |         addr.sin_family = AF_INET;
    |                           ^^^^^^^
    |
   ::: /.cargo/registry/src/github.com-1ecc6299db9ec823/libc-0.2.98/src/unix/linux_like/mod.rs:679:1
    |
679 | pub const AF_INET: ::c_int = 2;
    | ------------------------------- similarly named constant `_AF_INET` defined here
    |
help: a constant with a similar name exists
    |
65  |         addr.sin_family = _AF_INET;
    |                           ^^^^^^^^
help: consider importing this constant
    |
15  | use libc::AF_INET;
    |

error[E0412]: cannot find type `c_uint` in this scope
  --> /.cargo/registry/src/github.com-1ecc6299db9ec823/tun-0.5.1/src/platform/posix/sockaddr.rs:68:35
   |
68 |             s_addr: ((parts[3] as c_uint) << 24)
   |                                   ^^^^^^ not found in this scope
   |
help: consider importing one of these items
   |
15 | use libc::c_uint;
   |
15 | use std::os::raw::c_uint;
   |

error[E0412]: cannot find type `c_uint` in this scope
  --> /.cargo/registry/src/github.com-1ecc6299db9ec823/tun-0.5.1/src/platform/posix/sockaddr.rs:69:33
   |
69 |                 | ((parts[2] as c_uint) << 16)
   |                                 ^^^^^^ not found in this scope
   |
help: consider importing one of these items
   |
15 | use libc::c_uint;
   |
15 | use std::os::raw::c_uint;
   |

error[E0412]: cannot find type `c_uint` in this scope
  --> /.cargo/registry/src/github.com-1ecc6299db9ec823/tun-0.5.1/src/platform/posix/sockaddr.rs:70:33
   |
70 |                 | ((parts[1] as c_uint) << 8)
   |                                 ^^^^^^ not found in this scope
   |
help: consider importing one of these items
   |
15 | use libc::c_uint;
   |
15 | use std::os::raw::c_uint;
   |

error[E0412]: cannot find type `c_uint` in this scope
  --> /.cargo/registry/src/github.com-1ecc6299db9ec823/tun-0.5.1/src/platform/posix/sockaddr.rs:71:32
   |
71 |                 | (parts[0] as c_uint),
   |                                ^^^^^^ not found in this scope
   |
help: consider importing one of these items
   |
15 | use libc::c_uint;
   |
15 | use std::os::raw::c_uint;
   |

error[E0283]: type annotations needed
   --> /.cargo/registry/src/github.com-1ecc6299db9ec823/tun-0.5.1/src/configuration.rs:38:5
    |
35  | #[derive(Clone, Default, Debug)]
    |                 ------- in this derive macro expansion
...
38  |     pub(crate) platform: platform::Configuration,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type
    |
    = note: cannot satisfy `_: Default`
note: required by `std::default::Default::default`
   --> /.rustup/toolchains/nightly-x86_64-apple-darwin/lib/rustlib/src/rust/library/core/src/default.rs:116:5
    |
116 |     fn default() -> Self;
    |     ^^^^^^^^^^^^^^^^^^^^^
    = note: this error originates in the derive macro `Default` (in Nightly builds, run with -Z macro-backtrace for more info)

Some errors have detailed explanations: E0283, E0412, E0425, E0432.
For more information about an error, try `rustc --explain E0283`.
error: could not compile `tun` due to 10 previous errors
could not compile `tun` due to 10 previous errors
```

zonyitoo:
> --> /.cargo/registry/src/github.com-1ecc6299db9ec823/tun-0.5.1/src/lib.rs:28:9

Could you confirm that you have upgraded to the latest commit? The latest commit is now using the `tun` on crates.io.

zonyitoo:
Ok. I can reproduce that with Docker build.

zonyitoo:
Compilation finished on Docker. Please help and test on Android.

Mygod:
Cool! I'll take a look sometime.

Mygod:
Where do I start by the way? :)

dev4u:
感谢zonyitoo的快速响应，我也把代码合并到本地的shadowsock-android中做测试。

虽已成功编译，但是当前还不能正常工作。当中也许涉及有vpn service的代码修改：打开连接后，科学网络不通，tg连不上……

我猜测，sslocal建立的tun，与ss android建立的vpn都会接管所有的流量，他们在科学的时候，请求会先流经哪个？后流经的那个是做放行？还是会做什么操作？恐怕需要梳理一遍。

zonyitoo:
Start `sslocal` with `—tun-device-fd-from-path`, then it will listen to this path and get the fd from `VpnService`.

There is another requirement that the system must have a route setting that routes packets with destination to all IPs in the subnet to the

dev4u:
> Where do I start by the way? :)

- 合并ss rust的tun分支
- 修改core的build.gradle.kt，开启local-tun
- 修改启动sslocal的参数，增加: 
`--protocol tun --outbound-bind-interface lo0 --tun-interface-address 10.255.0.1/24`

zonyitoo:
On Android, the `sslocal` should get the device’s fd from VpnService instead of creating a new one.

zonyitoo:
I don’t know if Android could modify the route table in the Application. 

This route setting is required for TCP tunnel because in this implementation it requires this route setting to redirects packets from `tun` to a `TcpListener` that listens to the IP of the `tun` interface’s address.

dev4u:
> I don’t know if Android could modify the route table in the Application. 
> 
> This route setting is required for TCP tunnel because in this implementation it requires this route setting to redirects packets from `tun` to a `TcpListener` that listens to the IP of the `tun` interface’s address.

我印象不是很深，之前我浏览安卓那边的实现，好像在创建vpn对象时，是有设定路由规则。

iamwwc:
I have read all of codes about tun https://github.com/shadowsocks/shadowsocks-rust/blob/4287c2aab8f1826446b68888f58462b4342a0a53/crates/shadowsocks-service/src/local/tun/tcp.rs#L97

I found that you are implement a simple TCP protocol manually, as https://github.com/shadowsocks/shadowsocks-rust/issues/581#issuecomment-890502258 said, Have you consider ordering/retransmission into the situation ?

zonyitoo:
All TCP connections starts with a `SYN`. If retransmission happens, then it will just allocate a new source address and become a completely new connection. I don't think this is an issue here.

iamwwc:
What about packet reordering? If it occurred, local server will send wrong ordering packets to server

zonyitoo:
Packets are handled by a TcpListener (socket), which is a system provided TCP stack. So there is nothing have to be done in the Tun server.

zonyitoo:
Correct.

iamwwc:
On Linux, should we set a default route to routing all packets go through tun device manually? 

I don't see `routing on linux`  in README. only having a route setting related to MacOS

> (OPTIONAL) macOS requires adding a route entry to redirect packets that destinated to `10.155.0.1` (the address of the Tun interface) to the Tun interface itself.

zonyitoo:
Because I have tried on a Linux server, packets destinated to the tun's subnet would be routed properly.

iamwwc:
> Because I have tried on a Linux server, packets destinated to the tun's subnet would be routed properly.

I'm confused. If we don't set route table to routed all TCP/DNS packet to tun device, how we hijack TCP/DNS to give it a fake IP?

zonyitoo:
If you want to route all packets to the `tun` device, you will have to set the `default` route strategy.

The route setting mentioned in the README is for routing faked packets back to the `TcpListener`. It is the default behavior on Linux, so we don't need to do anything.

iamwwc:
> ref #581
> 
> ## Key design
> ### TCP
> image
> 
> 1. Create a `TcpListener` listening on an address in the subnet of `tun` device.
> 2. When `tun` receives TCP packets, it will redirect it to `TcpListener` by modifying `saddr` to a unique fake address and `daddr` to the address that `TcpListener` was listening.
> 3. `TcpListener` accepts a new connection. The connection's `peer_addr` will be the unique fake address. So we can get the original destination address by looking to the `tun`'s address allocation map.
> 
> ### UDP
> The same as the `redir` design, except it sends and receives packets from the `tun` interface.
> 
> ### Possible Applications
> * shadowsocks-android
> * shadowsocks-ios
> * macOS & Linux

But as above picture showed, we should set default route to route all client requests to tun device, so that we can assign a fake ip to destination.

I think tun protocol used for transparent proxy, which you don't set proxy settings manually in every application. Just set a default route, let tun device to handle packet.

> If you want to route all packets to the tun device, you will have to set the default route strategy.

If we don't set the default route, what the tun protocol used for?

zonyitoo:
Well, that's your choice. Setting the `default` route is not the only way to use `tun`.

For example, if you already have a tun device that controls all your packets, then you can simply set sslocal's tun as the outbound's interface.

iamwwc:
I guess I understand.

It like this, right?
```
          outgoing traffic
Surge tun ====> ss-tun
          SO_BINDTODEVICE

          Fake ip
Surge tun ====> ss-tun
```

zonyitoo:
```
ss-tun === fakeip ===> ss-tun
```

crates/shadowsocks-service/src/local/tun/tcp.rs
```diff
@@ -0,0 +1,272 @@
...
+        } = *(self.translator.lock().await);
+
+        let (conn, is_reply) = if tcp_header.syn && !tcp_header.ack {
+            // 1st SYN, creating a new connection
+            // Allocate a `saddr` for it
+            let saddr = loop {
+                let addr_idx = rand::random::<usize>() % self.free_addrs.len();
+                let port = rand::random::<u16>() % (65535 - 1024) + 1024;
+
+                let addr = SocketAddr::new(self.free_addrs[addr_idx], port);
```
iamwwc, line 104: Why we use random up to ten saddrs to determinate which saddr to used? I think use fixed one saddr is enough

iamwwc (commented):
Why we should pre allocate ten address for NAT?

crates/shadowsocks-service/src/local/tun/tcp.rs
zonyitoo: Well, then you can have at most `(65535 - 1024) * 10` free addresses.

iamwwc:
Does there will be having the windows tun support?

zonyitoo:
PRs are welcome. Since I don't have any Windows device, and there is no official tun interface support in Windows, I won't implement that.

iamwwc:
Will do after finish reading `shadowsocks-rust` codes. And I think the wintun project is a good choice

zonyitoo:
https://crates.io/crates/wintun

You may try to contribute it to https://crates.io/crates/tun .

crates/shadowsocks-service/src/local/tun/tcp.rs
zonyitoo: Now the `tun`'s address is included in the free addresses.

cirias:
Hi @dev4u Did you ever make this working on Android?

dev4u:
> Hi @dev4u Did you ever make this working on Android?

考虑到在使用体验上，tun与现在的vpn模式差别不大，所以我没有尝试在手机上tun。
如果你有兴趣，可以试试并欢迎分享你的结果。

PR #618, opened 2021-09-04
Update Dockerfile

juliatilde:
Running programs as root that don't require it is a big security risk. Docker is only containerization and not virtualization.
Instead of using the user "nobody" a custom user may be created, but as shadowsocks-rust only uses networking and only reads the config file I see it as unnecessary.
If the server is intended to bind on a privileged port (<1024) Docker's port mapping may be used or a program such as nftables or setcap may be used.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,6 +39,8 @@ COPY --from=build /root/shadowsocks-rust/target/release/sslocal /usr/bin
 
 COPY --from=build /root/shadowsocks-rust/examples/config.json /etc/shadowsocks-rust/
 
+USER nobody
+
 ENTRYPOINT [ "sslocal", "--log-without-time", "-c", "/etc/shadowsocks-rust/config.json" ]
 
 FROM alpine:3.14 AS ssserver
@@ -46,5 +48,7 @@ FROM alpine:3.14 AS ssserver
 COPY --from=build /root/shadowsocks-rust/target/release/ssserver /usr/bin
 
 COPY --from=build /root/shadowsocks-rust/examples/config.json /etc/shadowsocks-rust/
+
+USER nobody
 
 ENTRYPOINT [ "ssserver", "--log-without-time", "-c", "/etc/shadowsocks-rust/config.json" ]


kallydev (approved):
Yes, this is in line with the principle of least privilege, thank you. This modification has passed my test, it can now be merged. @zonyitoo

zonyitoo:
@kallydev The latest build were failed somehow.

juliatilde:
@zonyitoo I tried the Dockerfile on my Server (linux, amd64) before creating the PR and it appeared to have worked.
`docker build -t sstesting . --build-arg BUILDPLATFORM=linux --build-arg TARGETARCH=amd64` has returned
`Successfully built 7521edc99fdb`
`Successfully tagged sstesting:latest`
and when running it and then checking `netstat -tupnl` inside the container:
`tcp        0      0 127.0.0.1:8388          0.0.0.0:*               LISTEN      1/ssserver`
`whoami` also returned `nobody`.

zonyitoo:
https://github.com/shadowsocks/shadowsocks-rust/actions/runs/1203348443

Check this Action.

kallydev:
> https://github.com/shadowsocks/shadowsocks-rust/actions/runs/1203348443
> 
> Check this Action.

The reason for the error is that Docker tried to build the image based on a commit that does not exist in the master branch. The hash value of that commit is a0d91b51a15e6a2a1246c3c43544126c29388fcc.

This is probably because there was a forced push while GitHub Actions was building, and now we just need to rerun workflow.

kallydev:
> This is probably because there was a forced push while GitHub Actions was building, and now we just need to rerun workflow.

Rerun the workflow doesn't seem to solve the problem, because it has been bound to the corresponding commit.

If necessary, you can fork a local backup branch from master, then force master back to 58381e5470be21558ab38855a78c096ae9bccc50 and remove `v1.12.0-alpha.2` tag from remote. 

Finally, merge the backup branch to the local master, and repush the tag and commits to trigger the correct workflow.

PR #629, opened 2021-09-18
Use byte regexes to fix compiltation error

Y0ba:
Fixes error `error: pattern can match invalid UTF-8` in regular expressions. It still rejects unicode characters with another error `error: Unicode not allowed here` but that's fine since there is no unicode characters in dns records anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -637,7 +637,7 @@ Example configuration:
 # Match with subdomains
 ||google.com
 # An internationalized domain name should be converted to punycode
-# |☃-⌘\.com - WRONG
+# |☃-⌘.com - WRONG
 |xn----dqo34k.com
 # ||джpумлатест.bрфa - WRONG
 ||xn--p-8sbkgc5ag7bhce.xn--ba-lmcq
--- a/crates/shadowsocks-service/src/acl/mod.rs
+++ b/crates/shadowsocks-service/src/acl/mod.rs
@@ -13,7 +13,7 @@ use std::{
 
 use ipnet::{IpNet, Ipv4Net, Ipv6Net};
 use iprange::IpRange;
-use regex::{RegexSet, RegexSetBuilder};
+use regex::bytes::{RegexSet, RegexSetBuilder};
 
 use shadowsocks::{context::Context, relay::socks5::Address};
 
@@ -102,7 +102,7 @@ impl Rules {
 
     /// Check if the specified host matches any rules
     fn check_host_matched(&self, host: &str) -> bool {
-        self.rule_set.contains(host) || self.rule_tree.contains(host) || self.rule_regex.is_match(host)
+        self.rule_set.contains(host) || self.rule_tree.contains(host) || self.rule_regex.is_match(host.as_bytes())
     }
 
     /// Check if there are no rules for IP addresses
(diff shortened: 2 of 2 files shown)


zonyitoo:
I didn't notice that `Regex` has a byte only version. Great work!

Y0ba:
Should I add ascii check for `|asd.xyz` and `||asd.xyz` patterns? Now they accept international domains.

zonyitoo:
Good point. I think you should add it for consistency, the whole ACL file should have exactly one syntax constraint.

zonyitoo:
Just check every bytes of each lines if they are in `0..128`.

Y0ba:
String types have `is_ascii` too: https://doc.rust-lang.org/std/primitive.str.html#method.is_ascii

Y0ba:
Done. I also factored out some code into new struct `ParsingRules` to reduce code duplication and size of `load_from_file` function.

Y0ba:
@zonyitoo I also wonder about case sensitivity. Does ss' internal dns server lowercase dns requests? Maybe add lowercasing for `|asd.xyz` and `||asd.xyz` rules too? It's also possible to lowercase regex-patterns and disable case insensitivity in regex engine to improve performance event further.

zonyitoo:
Well, for socks5 and http proxy protocol, they didn't require domain names to be lowercase.

On the server side, the DNS client is using `trust-dns-resolver`, which I think will [translate names to lowercase](https://github.com/bluejekyll/trust-dns/blob/6dde9938d9f7541379b4307cb58a312607d230b5/crates/proto/src/rr/rdata/name.rs#L69) when sending queries (not sure).

zonyitoo:
Oh BTW, since ACLs are now limited to ASCII, should we convert IDNs to Punycode before checking if it matches any rules?

dev4u:
多谢小伙伴们快速发布了代码，修复了缺陷。

Y0ba:
I can convert `|...` and `||...`. But with regexes it's complicated (if not impossible), for example rule `一` is `xn--4gq` while `一一` is `xn--4gqa`, thus `一+` cannot be encoded in punycode in any meaningful way.

zonyitoo:
Should we convert domains to lowercase ASCII when handling `check_target_bypassed`?

Y0ba:
DNS server always operates with ASCII. There is no need to convert anything. And if `trust-dns-resolver` lowercases incoming dns requests there is no need to lowercase them too. But I wonder is there a way to add auto test for that?

Y0ba:
So, ss' DNS server already explicitly converts names to ascii. I can add here conversion to lowercase just in case.
https://github.com/shadowsocks/shadowsocks-rust/blob/731e613a60364712b45f4de23db38e1df23ee220/crates/shadowsocks-service/src/local/dns/server.rs#L313-L323

Y0ba:
> Should we convert domains to lowercase ASCII when handling `check_target_bypassed`?

Ooh. Now I see what did you mean. There are two places with acl checks. First is `check_name_in_proxy_list` and second is `check_target_bypassed`. `check_target_bypassed` doesn't related to dns and gets hosts directly from connection queries. Yea, here there should be both punycode and lowercase conversions.

Issue #569, opened 2021-07-09
Any plan to release arm-apple-darwin binaries for M1 Mac and future Apple silicon devices?


ruixingw:
The title explains.

zonyitoo:
There is no way to cross compile it from Github Action yet. Feel free to make a PR if you find a way.

ruixingw:
Oh I see.
I'll follow this issue here: https://github.com/actions/virtual-environments/issues/2187.

zonyitoo:
The target should be `aarch64-apple-darwin` for M1 devices.

ruixingw:
> The target should be `aarch64-apple-darwin` for M1 devices.

I know it's the same thing but it seems that Apple calls it `arm64` instead of `aarch64`:
```
$ uname -a
Darwin <devicename> 20.5.0 Darwin Kernel Version 20.5.0: Sat May  8 05:10:31 PDT 2021; root:xnu-7195.121.3~9/RELEASE_ARM64_T8101 arm64
```

zonyitoo:
https://doc.rust-lang.org/nightly/rustc/platform-support.html

Nope. It is `aarch64` in LLVM (rustc)'s target.

PR #634, opened 2021-09-24
CI: build check and release for Apple Silicon

deantvv:
Currently we can only use x86_64 macOS to cross-build aarch64 platform, which is this PR try to do.

For #569

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -8,7 +8,6 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
-  RUSTFLAGS: "-Ctarget-feature=+aes,+ssse3"
   RUST_LOG: "trace"
 
 jobs:
@@ -61,3 +60,21 @@ jobs:
             -W clippy::absurd_extreme_comparisons
             -W clippy::erasing_op
             -A clippy::collapsible_else_if
+      - name: prepare AppleSilicon build
+        uses: actions-rs/toolchain@v1
+        with:
+          target: aarch64-apple-darwin
+        if: runner.os == 'macOS'
+
+      # https://github.com/actions/virtual-environments/issues/2557#issuecomment-769611326
+      - if: runner.os == 'macOS'
+        run: |
+          sudo xcode-select -s /Applications/Xcode_12.4.app &&
+          sudo rm -Rf /Library/Developer/CommandLineTools/SDKs/*
+
+      - name: Check compilation for Apple Silicon
+        uses: actions-rs/cargo@v1
+        with:
+          command: check
+          args: --target aarch64-apple-darwin
+        if: runner.os == 'macOS'
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -68,6 +68,9 @@ jobs:
       matrix:
         # os: [ubuntu-latest, macos-latest]
         os: [macos-latest]
+        target:
+          - x86_64-apple-darwin
+          - aarch64-apple-darwin
     steps:
       - uses: actions/checkout@v2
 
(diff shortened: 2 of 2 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -61,3 +61,23 @@ jobs:
...
+      # https://github.com/actions/virtual-environments/issues/2557#issuecomment-769611326
+      - if: runner.os == 'macOS'
+        run: |
+          sudo xcode-select -s /Applications/Xcode_12.4.app &&
+          sudo rm -Rf /Library/Developer/CommandLineTools/SDKs/*
+
+      - name: Check compilation for Apple Silicon
+        uses: actions-rs/cargo@v1
+        env:
+          RUSTFLAGS: "-Ctarget-feature=+aes"  # aarch64-apple-darwin does not support ssse3
```
zonyitoo, line 79: It is Ok to remove `RUSTFLAGS` completely because thay have already been defined in `.cargo/config.toml`.

.github/workflows/build-release.yml
```diff
@@ -83,11 +86,17 @@ jobs:
         with:
           profile: minimal
 
+      # https://github.com/actions/virtual-environments/issues/2557#issuecomment-769611326
+      - if: ${{ matrix.target }} == 'aarch-apple-darwin'
+        run: |
+          sudo xcode-select -s /Applications/Xcode_12.4.app &&
```
zonyitoo, line 92: Is that `&&` mandatory?

.github/workflows/build-and-test.yml
zonyitoo: Remove the `RUSTFLAGS` in the global scope.

.github/workflows/build-release.yml
deantvv: Yes, it is. Because github action runner will report the status code of the last line, we need to make sure this is executed as a single line.

[ref](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstepsrun)
> sh-like shells exit with the exit code of the last command executed in a script, which is also the default behavior for actions. The runner will report the status of the step as fail/succeed based on this exit code.

.github/workflows/build-and-test.yml
deantvv: Removed `RUSTFLAGS` as suggested.

zonyitoo (approved):

zonyitoo:
You may have to improve the release build script: https://github.com/shadowsocks/shadowsocks-rust/runs/3708914876?check_suite_focus=true

deantvv:
@zonyitoo Could you explain a little bit more about the possible improvement?

zonyitoo:
```
'+pmull' is not a recognized feature for this target (ignoring feature)
error[E0463]: can't find crate for `core`
  |
  = note: the `aarch64-apple-darwin` target may not be installed
  = help: consider downloading the target with `rustup target add aarch64-apple-darwin`
  = help: consider building the standard library from source with `cargo build -Zbuild-std`
```

It fails to build. You may have to install the toolchain first.

deantvv:
> 
> 
> ```
> '+pmull' is not a recognized feature for this target (ignoring feature)
> error[E0463]: can't find crate for `core`
>   |
>   = note: the `aarch64-apple-darwin` target may not be installed
>   = help: consider downloading the target with `rustup target add aarch64-apple-darwin`
>   = help: consider building the standard library from source with `cargo build -Zbuild-std`
> ```
> 
> It fails to build. You may have to install the toolchain first.

Thanks. I will look more into this.

zonyitoo:
You may start from adding a `target` in here

https://github.com/shadowsocks/shadowsocks-rust/blob/049c5c43748f9bc65a0d75744b0cab9b5f564727/.github/workflows/build-release.yml#L87

PR #635, opened 2021-09-26
ci: Fix build release

deantvv:
Fix for #634


zonyitoo:
It may have to install Rust nightly because Aarch64 requires unstable features (SIMD).

PR #662, opened 2021-10-31
Add v2ray-plugin Docker image example

realies:
If this is to go into the Docker image release workflow, it needs to pull the correct architecture version of the plugin.

### DIFF
--- a/Dockerfile.v2ray
+++ b/Dockerfile.v2ray
@@ -1,0 +1,15 @@
+FROM ghcr.io/shadowsocks/ssserver-rust:latest
+
+USER root
+
+RUN cd /tmp && \
+ TAG=$(wget -qO- https://api.github.com/repos/shadowsocks/v2ray-plugin/releases/latest | grep tag_name | cut -d '"' -f4) && \
+ wget https://github.com/shadowsocks/v2ray-plugin/releases/download/$TAG/v2ray-plugin-linux-amd64-$TAG.tar.gz && \
+ tar -xf *.gz && \
+ rm *.gz && \
+ mv v2ray* /usr/bin/v2ray-plugin && \
+ chmod +x /usr/bin/v2ray-plugin
+
+USER nobody
+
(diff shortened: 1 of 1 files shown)


realies:
@zonyitoo, should the v2ray-plugin be added in the current Docker image or as a separate tag?

zonyitoo:
It shouldn't be included in this Project. If `v2ray-plugin` could be packaged in the image, why wouldn't the others?

realies:
@zonyitoo, primarily because the v2ray-plugin is maintained by the shadowsocks organisation. It would be useful to have an official image that includes it, would the v2ray-plugin repository be more suitable?

zonyitoo:
I think you could modify this Dockerfile to `Dockerfile.plugin` which could includes all official maintained plugins.

But, is it possible to make a `v2ray-plugin` only image and then run multiple images in a k8s group? Then you won't need to package all of them into one image.

realies:
@zonyitoo, Is there a list of all plugins maintained by shadowsocks somewhere? https://shadowsocks.org/en/wiki/Plugin.html points me to kcptun and v2ray-plugin.

Not sure I understand the question about the k8s group - assume you mean creating a plugin image and mounting its binary into a barebones shadowsocks-rust container? Not aware if that's possible, but if it is, it may require Kubernetes.

zonyitoo:
You could just create a `Dockerfile.plugin` file for creating an image containing `sslocal` or `ssserver` with plugins. And then make a document for guiding how to add new plugins into this image.

All the following requests about adding plugins could make a PR to add it into this `Dockerfile.plugin`.

This solution have some obvious flaws:

1. The image might be large because it will contain many plugins that you don't need.
2. When you need a new plugin, making a new PR, merge, waiting for the next release may be too long for users that need to use it immediately.

Kubernetes should be a possible solution.

Based on your description (about mounting binaries), there should be a solution that you could just download the `v2ray-plugin` releases on your host machine and then mount them into the container's `/usr/bin/v2ray-*` paths. Flexible, and easy to manage. One possible flaw is that if the plugin requires some other runtime, such as `Python`, then it will become an issue.

Issue #669, opened 2021-11-05
Dealing with unicode domains in ACL


spyophobia:
Continues discussion under 4500d46

Y0ba:
Also in your particular example all three rules can be replaced with this:
```
# .中国
||xn--fiqs8s
# .公司
||xn--55qx5d
# .网络
||xn--io0a7i
```

spyophobia:
> Because dns in general doesn't work with unicode domains. It's possible to do it in opposite way - convert everything to unicode and then do match, but I not sure if it worth it.

Okay I can respect this opinion. Also after a bit of research I found the standard for this (rfc3492) so I guess yeah you're correct.

However I am more concerned with backward compatibility. Many ACLs exist that are not properly encoded considering this isn't exactly a well-known standard. Also many users simply pull existing ACLs from others' repositories instead of writing their own, which means they rely on the upstream to fix their lists to prevent annoying breakages.

For this reason I suggest that we at minimum introduce an extensive deprecation period for unencoded Unicode domains.  Again, if we cannot match against unicode patterns, at least don't let one bad line fail the entire parse (unless there are any associated safety concerns).

Y0ba:
Okay, here is what I think is the best option here:
1) Enable unicode support and explicitly convert all domains to unicode (ind converter is smart and don't convert ascii domains so there won't be a performance impact)/
2) Detect `asd$` and `(^|\.)xyz$` (also `(?:^|\.)xyz$`) patterns and automatically replace them with `|asd` and `||xyz`.

That way we will get both high speed (speed of `|xyz` and `||xyz` is unaffected by unicode) and  backward compatibility. It will be even faster since 99% of regex rules are exactly these two patterns.

spyophobia:
I'm not very familiar with IDNs at all (or for that matter, the inner workings of ACL; I simply noticed a breakage in my ACL after update lol). So I will defer to your opinion on the specific implementation.

Y0ba:
Well, unicode conversion was introduced not because of some some principle, but simply because unicode regexes are slower and take more memory. If this slowdown can be mitigated for most of regexes, then there is no point to not support idns.

zonyitoo:
What error did you got? Could you provide an ACL file that fail to parse? @spyophobia

dev4u:
https://github.com/shadowsocks/shadowsocks-rust/pull/627#issuecomment-922320750
[...]

PR #673, opened 2021-11-08
ACL regex rules convert to tree-rules or set-rules

zonyitoo:
ref #669

### DIFF
--- a/crates/shadowsocks-service/src/acl/mod.rs
+++ b/crates/shadowsocks-service/src/acl/mod.rs
@@ -10,11 +10,14 @@ use std::{
     io::{self, BufRead, BufReader, Error, ErrorKind},
     net::{IpAddr, SocketAddr},
     path::Path,
+    str,
 };
 
 use ipnet::{IpNet, Ipv4Net, Ipv6Net};
 use iprange::IpRange;
-use regex::bytes::{RegexSet, RegexSetBuilder};
+use log::trace;
+use once_cell::sync::Lazy;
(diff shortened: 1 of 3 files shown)


zonyitoo:
@Y0ba Request For Comments.

Y0ba:
You can actually merge both rules into one regex, like this: `^(?:(?:\((?:\?:)?\^\|\\\.\)|(?:\^\.\+)?\\\.)?((?:[\w-]+(?:\\\.)?)+)|\^((?:[\w-]+(?:\\\.)?)+))\$$`. Then if `caps.get(1)` returns `Some(..)` it's a tree-like pattern, if `caps.get(2)` returns `Some(...)` - list-like pattern.

crates/shadowsocks-service/src/acl/mod.rs
```diff
@@ -158,25 +161,76 @@ impl ParsingRules {
...
+                if let Ok(tree_rule) = str::from_utf8(tree_rule.as_bytes()) {
+                    let tree_rule = tree_rule.replace("\\.", ".");
+                    if let Ok(..) = self.add_tree_rule_inner(&tree_rule) {
+                        trace!("REGEX-RULE {} => TREE-RULE {}", rule, tree_rule);
+                        return;
+                    }
+                }
+            }
+
+            if let Some(set_rule) = caps.get(2) {
```
Y0ba, line 196:
```suggestion
            } else if let Some(set_rule) = caps.get(2) {
```
Since it's not possible for both rules match at the same time you can use else if here.

Y0ba (approved):

PR #701, opened 2021-11-27
Local, Server & Manager unified binary ssservice

### LINKED ISSUE #698: Suggest: Build one single bundled binary
For proxy tools, source code that are required by client **is** also required by server, in theory.
sslocal, ssserver, ssmanager shares a lot of common source code.
Static-link build them apart is wasting storage space.As of ssurl, it's something 'one line code'.
I suggest build them into one single binary file, with switch flag, like --local, --server, --manager, --encode, --decode.

zonyitoo:
- fixes #698
- Refactored shadowsocks-rust crate structure. Common code are put into a separated library


database64128:
What about putting `ssserver` in a separate zip file? The user can then choose which combination of binaries to download.

zonyitoo:
Do you mean `ssservice`? Well, network download speed is fast enough in 2021. Just keep it until there are more users complain about it.

Issue #739, opened 2022-01-04
proposal(deploy): support k8s cloud native deployment


longkai:
I've been deploying shadowsocks in Kubernetes with a cloud provider or bare metal node for a while, and its deployment experience is really fine.

What's more, we can leverage k8s Service to expose traffic outside, like `LoadBalancer` or `NodePort` which gains more fine-grained compare with fixed host and port.

For a more interesting use case, we can use a Ingress(Istio, nginx, etc.) which routes the matched traffic to shadowsocks along with real web service.

I've written a helm chart with necessary values that user can modify(method, password, etd.), then `helm install` all the stuff into k8s with one click.

Or just generate the raw yaml then `kubectl apply -f shadowsocks.yaml`.

Since the cloud native is more popular, if you think this k8s deployment is fine, I would be happy send a PR.

Thank you for your nice work.

kallydev:
This usage is very cool, I am very interested in it.

longkai:
> This usage is very cool, I'm very interested in it.
> 
> Is it based on the Docker image we provide at GitHub Packages?

Yes.

PR #741, opened 2022-01-09
feat(deploy): support k8s deployment (#739)

longkai:
- helm chart
- kubectl with raw manifest which generated via helm
- update README

One more thing, we could host the chart repo with GitHub pages or within a Docker oci image.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,66 @@ docker run --name ssserver-rust \
   -dit ghcr.io/shadowsocks/ssserver-rust:latest
 ```
 
+### **Deploy to Kubernetes**
+
+This project provided yaml manifests for deploying to Kubernetes.
+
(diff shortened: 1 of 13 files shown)


zonyitoo:
Do we need to change the version in all configuration files every time when we are making a new release?

zonyitoo:
On the other hand, you are using a `manifest` folder in the Project root, which is implicit for users to realize that this is for k8s (helm). Could you put them in a `k8s/manifest` or `helm/manifest`?

longkai:
The app version of the helm chart is unnecessary to change each release if there is no breaking changes because users can always change the docker image if they want(default to latest).

I suppose the right time to change version is when we change the chart itself.

For more clarity I've marked the app version to `1.x.x`.

And the folder name is change to `k8s`.

PR #742, opened 2022-01-09
chore(k8s): change default image to ghcr latest

longkai:
Sorry for the last PR. I forgot to change the default docker image after local testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,16 +154,14 @@ Below is the common default values you can change:
 
 ```yaml
 # This is the shadowsocks config which will be mount to /etc/shadowocks-rust.
-# You can put arbitrary yaml here, it will translation to json before mounting.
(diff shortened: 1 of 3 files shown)


zonyitoo:
And you should delete the `nameserver` key in the configuration file which is not recognized in this project.

longkai:
Done

database64128:
Might as well remove the `timeout` field and change the listening address to `::`. Unlike Go, Rust programs generally don't interpret `0.0.0.0` as `::`.

zonyitoo:
Agree. Listening on `::` with dual-stack IPv6 support should cover more users' use cases.

longkai:
Those fields comes from libev and I've done that.

BTW, I still see some timeout and 0.0.0.0 server address in the project, but it's better with another dedicated PR.

PR #750, opened 2022-01-22
Strip symbols in release mode

### LINKED ISSUE #276: Would you please strip for release assets?
I think you can strip for release assets so that thin the tar package.

Y0ba:
Adds strip option since it was [stabilized](https://github.com/rust-lang/cargo/pull/10088). Closes #276.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,6 +50,7 @@ lto = "fat"
 codegen-units = 1
 incremental = false
 panic = "abort"
+strip = true
 
 [features]
 default = [


zonyitoo:
BTW, should we use `strip = "symbols"` instead of just `strip = true`?

Y0ba:
It's same. Per documentation:
> You can also configure this option with the two absolute boolean values
> `true` and `false`. The former enables `strip` at its higher level, `symbols`

You can change it to "symbols" if you want to be more explicit.

PR #751, opened 2022-01-22
Disable secure mode in mimalloc allocator

Y0ba:
By default mimalloc has ["secure"](https://github.com/microsoft/mimalloc#secure-mode) feature enabled which has some performance penalty and a little memory overhead (due to guard pages). It's useless for safe rust code, but can be useful for unsafe. I'm not sure if it useful for shadowsocks-rust project, but personally I disable it in my builds.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -153,7 +153,7 @@ tokio = { version = "1", features = ["rt", "signal"] }
 
 ipnet = { version = "2.3", optional = true }
 
-mimalloc = { version = "0.1", optional = true }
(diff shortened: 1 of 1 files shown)


zonyitoo:
What platform are you running with mimalloc? It doesn't work on OpenWRT with MUSL build.

Y0ba:
I use it with amd64+musl (within docker).

Btw arm+musl problems seems like are already [fixed](https://github.com/microsoft/mimalloc/issues/495) in unstable `dev` version.

PR #774, opened 2022-02-25
remove deprecated code for clap 3.1.0

18o:


zonyitoo:
Then you should set the version to `3.1`:

https://github.com/shadowsocks/shadowsocks-rust/blob/2c85eef5181a2df049d2073fb4799545d60157eb/Cargo.toml#L147

18o:
I saw the clap version in Cargo.lock has already set to 3.1 by dependabot, does the Cargo.toml still need update manually ?

zonyitoo:
Yes it should be updated. Your modification won't be able to compile with clap < 3.1

PR #821, opened 2022-04-26
Update build-release.yml to add popular armv7 build targets

xiazuojie:
add popular armv7 build targets


zonyitoo:
You can only add targets that are supported here: https://github.com/cross-rs/cross#supported-targets .

You should only add the target that are useful for most people. It is unwise to add all the targets wasting the release time.

xiazuojie:
Changed to only include the following 2.
* armv7-unknown-linux-musleabihf (static binary)
* armv7-unknown-linux-gnueabihf (dynamic linked)

PR #844, opened 2022-05-22
update: remove rnd from security-iv-printable-prefix feature

chuxi:
I wrote little rust. So it is welcome to tell me any problems in the commit.

The feature `security-iv-printable-prefix` works well for me. I found the original solution used `Rand` package to generate the index. It is repeated for nonce calculation. So I updated the codes to prefix processing of nonce.


zonyitoo:
Clever move.

zonyitoo:
Your test `fn generate_nonce()` fails to compile.

chuxi:
yes. I tested it on v1.14.3. So missed the test on master branch. just fixed it.

zonyitoo (approved):

PR #985, opened 2022-10-25
fix(ssurl): use `percent_decode_str` for `userinfo` encoded data

RAX7:
Some URLs contain a base64 string with `=` char (e.g: `ss://YWVzLTEyOC1nY206c2hhZG93c29ja3M=@1.2.3.4:8000#tag`) which breaks the `ssurl` tool.

### DIFF
--- a/crates/shadowsocks/src/config.rs
+++ b/crates/shadowsocks/src/config.rs
@@ -12,7 +12,7 @@ use std::{
     time::Duration,
 };
 
-use base64::{decode_config, encode_config, URL_SAFE_NO_PAD};
+use base64::{decode_config, encode_config, URL_SAFE, URL_SAFE_NO_PAD};
 use byte_string::ByteStr;
 use bytes::Bytes;
 use cfg_if::cfg_if;
@@ -668,6 +668,9 @@ impl ServerConfig {
                 (m, p)
             }
             None => {
+                // userinfo is not required to be percent encoded, but some implementation did.
+                // If the base64 library have padding = added to the encoded string, then it will become %3D.
+
                 let decoded_user_info = match percent_encoding::percent_decode_str(user_info).decode_utf8() {
                     Ok(m) => m,
                     Err(err) => {
@@ -676,7 +679,18 @@ impl ServerConfig {
                     }
                 };
 
-                let account = match decode_config(decoded_user_info.to_string(), URL_SAFE_NO_PAD) {
+                // reborrow to fit AsRef<[u8]>
+                let decoded_user_info: &str = &decoded_user_info;
+
+                let base64_config = if decoded_user_info.ends_with('=') {
+                    // Some implementation, like outline,
+                    // or those with Python (base64 in Python will still have '=' padding for URL safe encode)
+                    URL_SAFE
+                } else {
+                    URL_SAFE_NO_PAD
+                };
+
+                let account = match decode_config(decoded_user_info, base64_config) {
                     Ok(account) => match String::from_utf8(account) {
                         Ok(ac) => ac,
                         Err(..) => return Err(UrlParseError::InvalidAuthInfo),
--- a/crates/shadowsocks/src/config.rs
+++ b/crates/shadowsocks/src/config.rs
@@ -668,7 +668,15 @@ impl ServerConfig {
                 (m, p)
             }
             None => {
-                let account = match decode_config(user_info, URL_SAFE_NO_PAD) {
+                let decoded_user_info = match percent_encoding::percent_decode_str(user_info).decode_utf8() {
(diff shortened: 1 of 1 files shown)


zonyitoo:
This is impossible. The userinfo has to be encoded with base64 url safe encoding, which is definitely don't have any `=`. `ss://YWVzLTEyOC1nY206c2hhZG93c29ja3M=@1.2.3.4:8000#tag` is not a valid SIP002 URL.

RAX7:
@zonyitoo base64 may contain `=` for padding
> = padding characters might be added to make the last encoded block contain four Base64 characters.
> https://en.wikipedia.org/wiki/Base64

zonyitoo:
You are messing up "base64" and "base64 url safe" encoding.

1. Read the "URL Applications" section in your provided wiki: https://en.wikipedia.org/wiki/Base64#URL_applications
2. RFC4648 base64url: https://datatracker.ietf.org/doc/html/rfc4648#section-5

database64128:
Your example is an invalid URL with characters outside the allowed set but not percent encoded. This won't even work with some URL parsers. I don't know how you obtained this "URL", but you should tell whoever generated it to use a proper URL builder, not string concatenation.

RAX7:
Oh yes you are right.
> You should make a PR to fix where the URL was generated.

I can't do that, I'm just trying to use [outline](https://getoutline.me/access-keys/8856/) keys and `ssurl` breaks with error.
May I create another PR witch remove all `=` chars from userinfo to make this tool more user friendly?

zonyitoo:
Well, the code in PR actually trying to decode the userinfo with percent encoding, so the actual example should be: `ss://YWVzLTEyOC1nY206c2hhZG93c29ja3M%3D@1.2.3.4:8000#tag`

zonyitoo:
> May I create another PR witch remove all `=` chars from userinfo to make this tool more user friendly?

No you should not do that. You may also need to check if there is any characters that are not in the valid base64 url safe character set and try to fix it to fulfill the specification requirement.

You will never be able to compatible all kinds of errors, which is why we need a specification.

But on the otherhand, I noticed that some base64 encoder, such as [the one in Python](https://docs.python.org/3/library/base64.html#base64.urlsafe_b64encode) will still add tailing `=`s for url safe encoding. So I think this PR is still valid for this situation.

database64128:
https://github.com/Jigsaw-Code/outline-shadowsocksconfig/blob/master/src/shadowsocks_config.ts#L338

It seems they are indeed using string concatenation. You should also open an issue over there.

zonyitoo:
```typescript
const userInfo = Base64.encodeURI(`${method.data}:${password.data}`);
```

https://github.com/dankogai/js-base64/blob/68f12dea403e8b0031b09fa5edd0904066fb7ab4/base64.ts#L121-L133

```typescript
const _mkUriSafe = (src: string) => src
    .replace(/=/g, '').replace(/[+\/]/g, (m0) => m0 == '+' ? '-' : '_');

/**
 * converts a UTF-8-encoded string to a Base64 string.
 * @param {boolean} [urlsafe] if `true` make the result URL-safe
 * @returns {string} Base64 string
 */
const encode = (src: string, urlsafe = false) => urlsafe
    ? _mkUriSafe(_encode(src))
    : _encode(src);
/**
 * converts a UTF-8-encoded string to URL-safe Base64 RFC4648 §5.
 * @returns {string} Base64 string
 */
const encodeURI = (src: string) => encode(src, true);
```

So the outline generated URL shouldn't contain any `=`.

RAX7:
@zonyitoo the original URL I took [here](https://getoutline.me/access-keys/8856/) and it's: `ss://Y2hhY2hhMjAtaWV0Zi1wb2x5MTMwNTpHIXlCd1BXSDNWYW8=@ak1489.free.www.outline.network:806#www.outline.network%20(canada)`

zonyitoo:
Interesting, the code in outline-server is Ok but the console is running a different code version.

Please checkout the comments in the review, I am Ok to merge this PR if they are fixed.

RAX7:
@zonyitoo, @database64128 I guess they using `outline-shadowsocksconfig` only for desktop client, but not for website
https://github.com/Jigsaw-Code/outline-client/blob/master/package.json#L54

> Please checkout the comments in the review, I am Ok to merge this PR if they are fixed.

Test cases fails on windows with:
```
Error: There is not enough space on the disk. : 'D:\a\_temp\_runner_file_commands\step_summary_7596271d-8edb-440d-9476-a7e3c69f2199'
```

zonyitoo:
Just ignore the Windows test case.

zonyitoo (changes requested):

PR #987, opened 2022-10-28
Implemented poll_* family functions on UDP ProxySocket

ibigbug:
so that it allows the socket to be wrapped into Stream/Sink easily

### DIFF
--- a/crates/shadowsocks/src/relay/socks5.rs
+++ b/crates/shadowsocks/src/relay/socks5.rs
@@ -16,7 +16,10 @@ use bytes::{Buf, BufMut, BytesMut};
 use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
 
 pub use self::consts::{
-    SOCKS5_AUTH_METHOD_GSSAPI, SOCKS5_AUTH_METHOD_NONE, SOCKS5_AUTH_METHOD_NOT_ACCEPTABLE, SOCKS5_AUTH_METHOD_PASSWORD,
+    SOCKS5_AUTH_METHOD_GSSAPI,
+    SOCKS5_AUTH_METHOD_NONE,
+    SOCKS5_AUTH_METHOD_NOT_ACCEPTABLE,
+    SOCKS5_AUTH_METHOD_PASSWORD,
 };
 
 #[rustfmt::skip]
--- a/crates/shadowsocks/src/relay/socks5.rs
+++ b/crates/shadowsocks/src/relay/socks5.rs
@@ -16,10 +16,7 @@ use bytes::{Buf, BufMut, BytesMut};
 use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
 
 pub use self::consts::{
-    SOCKS5_AUTH_METHOD_GSSAPI,
-    SOCKS5_AUTH_METHOD_NONE,
-    SOCKS5_AUTH_METHOD_NOT_ACCEPTABLE,
-    SOCKS5_AUTH_METHOD_PASSWORD,
+    SOCKS5_AUTH_METHOD_GSSAPI, SOCKS5_AUTH_METHOD_NONE, SOCKS5_AUTH_METHOD_NOT_ACCEPTABLE, SOCKS5_AUTH_METHOD_PASSWORD,
 };
 
 #[rustfmt::skip]
@@ -247,7 +244,7 @@ impl Address {
                 let mut buf = vec![0u8; domain_len];
                 cur.copy_to_slice(&mut buf);
                 let port = cur.get_u16();
-                let addr = String::from_utf8(buf).map_err(|x| Error::AddressDomainInvalidEncoding)?;
+                let addr = String::from_utf8(buf).map_err(|_| Error::AddressDomainInvalidEncoding)?;
                 Ok(Address::DomainNameAddress(addr, port))
             }
             _ => Err(Error::AddressTypeNotSupported(atyp)),
--- a/crates/shadowsocks/src/relay/socks5.rs
+++ b/crates/shadowsocks/src/relay/socks5.rs
@@ -12,14 +12,11 @@ use std::{
     vec,
 };
 
-use bytes::{BufMut, BytesMut};
+use bytes::{Buf, BufMut, BytesMut};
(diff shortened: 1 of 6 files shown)


zonyitoo:
You should use the nightly toolchain and rerun `cargo fmt`. The code format is broken.

crates/shadowsocks/src/relay/socks5.rs
```diff
@@ -213,6 +210,47 @@ pub enum Address {
 }
 
 impl Address {
+    /// read from a cursor
+    pub fn read_cursor<T: AsRef<[u8]>>(cur: &mut io::Cursor<T>) -> Result<Address, Error> {
```
zonyitoo, line 214: You just made a new function `Address::read_cursor`, then it will have exactly the same code as the `Address::read_from`. Is it possible to unify these two functions?

crates/shadowsocks/src/relay/udprelay/proxy_socket.rs
```diff
@@ -254,6 +251,76 @@ impl ProxySocket {
         Ok(send_len)
     }
 
+    pub fn poll_send(&self, addr: &Address, payload: &[u8], cx: &mut Context<'_>) -> Poll<ProxySocketResult<usize>> {
```
zonyitoo, line 254: Since you have moved the main logic from `send` to `poll_send`, then you should use `futures::future::poll_fn` to reimplement `send` with `poll_send`.

crates/shadowsocks/src/relay/udprelay/proxy_socket.rs
```diff
@@ -254,6 +251,76 @@ impl ProxySocket {
...
+                    Poll::Ready(Ok(payload.len()))
+                } else {
+                    Poll::Ready(Err(io::Error::from(ErrorKind::WriteZero).into()))
+                }
+            }
+            x => x,
+        }
+    }
+
+    pub fn poll_send_to(
```
zonyitoo, line 279: `poll_send_to_with_ctrl` and `poll_send_with_ctrl` are missing.

crates/shadowsocks/src/relay/udprelay/proxy_socket.rs
```diff
@@ -376,6 +440,40 @@ impl ProxySocket {
             .map(|(n, sa, a, rn, _)| (n, sa, a, rn))
     }
 
+    /// poll family functions
+    pub fn poll_recv(
```
zonyitoo, line 444: same as the `send`s function.

zonyitoo (changes requested):

crates/shadowsocks/src/relay/socks5.rs
ibigbug: tbh i'm not sure - the logic are same, but reusing the sync and async calls seems non trivial. you got any suggestions?

crates/shadowsocks/src/relay/udprelay/proxy_socket.rs
ibigbug: i think send and poll send are essentially different family function calls which should be calling each other internally.

as inside the send call, timeout is taken care of while the poll_send doesn't have that logic and just return different poll state.

wdyt?

crates/shadowsocks/src/relay/udprelay/proxy_socket.rs
ibigbug: updated

crates/shadowsocks/src/relay/udprelay/proxy_socket.rs
ibigbug: tokio also seems to implmement these two separately https://docs.rs/tokio/latest/src/tokio/net/udp.rs.html#768-773

zonyitoo: > as inside the send call, timeout is taken care of

Hmm, I don't think so. Async calls doesn't have timeouts.

crates/shadowsocks/src/relay/socks5.rs
zonyitoo: You are right. I am ok with it.

crates/shadowsocks/src/relay/udprelay/proxy_socket.rs
zonyitoo: https://github.com/tokio-rs/tokio/blob/a66884a2fb80d1180451706f3c3e006a3fdcb036/tokio/src/net/tcp/split.rs#L137-L140

I am not really sure why tokio implements UdpSocket's `recv` separately. But some part of its code actually uses `poll_fn` to reuse `poll_*`s.

ibigbug: right i was more referring to this implementation detail:

https://github.com/shadowsocks/shadowsocks-rust/blob/ab6a6f6178d3a29e9eea16fd73b21c80a7a77278/crates/shadowsocks/src/relay/udprelay/proxy_socket.rs#L340-L347

it would a bit tricky to check timeout on the poll_fn, I suppose?

ibigbug: i think if it's just a simple wrap it might be fine, otherwise maybe implement different logic separately

zonyitoo: Ah you are right, I just forgot there is a `self.recv_timeout`. Well yes, it will be very tricky to implement timeout in `poll_recv`. You should mention that in the doc.

ibigbug: or just simply 

```rust
let recv_n = match self.recv_timeout { 
     None => poll_fn(...).await?, 
     Some(d) => match time::timeout(d, poll_fn(...).await? { 
         Ok(Ok(l)) => l, 
         Ok(Err(err)) => return Err(err.into()), 
         Err(..) => return Err(io::Error::from(ErrorKind::TimedOut).into()), 
     }, 
 }; 
```

i'm happy to go with either way

zonyitoo: I am ok with your simple solution.

zonyitoo:
CI failed.

zonyitoo (approved):

PR #1082, opened 2023-01-13
bugfix: salt recheck will cause duplicated iv/salt error

chuxi:
in version 1.14.3

```
        // Check repeated salt after first successful decryption #442
        if self.salt.is_some() {
            let salt = self.salt.take().unwrap();
            context.check_nonce_replay(&salt)?;
        }
```

current master version

```
        // Check repeated salt after first successful decryption #442
        if let Some(ref salt) = self.salt {
            context.check_nonce_replay(self.method, salt)?;
        }
```

if enable feature `security-replay-attack-detect`, it causes error `detected repeated nonce (iv/salt)`


chuxi:
we need to revert the code, or aead can not work in current version if we set reject policy.

PR #1086, opened 2023-01-14
Deal with warnings caused by outdated actions in workflows

zedifen:
Currently the workflow is using `actions/checkout@v2`, which is outdated and generates warnings in workflow logs. This PR upgrade `actions/checkout` to the latest `v3`.

There are also two actions `actions-rs/toolchain` and `actions-rs/clippy-check` from [actions-rs](https://github.com/actions-rs/) in use, which is also outdated and haven't been updated in years.

This PR replaces `actions-rs/toolchain` with bare `rustup` commands, since `rustup` is included in GitHub Actions runner images (see [actions/runner-images](https://github.com/actions/runner-images)).

However there seems to be no easy way for replacing `actions-rs/clippy-check`. So, I'm marking this a draft PR and open for further discussions.

### DIFF
--- a/.github/workflows/clippy-check.yml
+++ b/.github/workflows/clippy-check.yml
@@ -30,7 +30,7 @@ jobs:
           rustup default nightly
           rustup override set nightly
       - name: Clippy Check
-        uses: actions-rs/clippy-check@v1
+        uses: kristof-mattei/clippy-check@main
         with:
           name: clippy-${{ matrix.platform }}
           token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/clippy-check.yml
+++ b/.github/workflows/clippy-check.yml
@@ -30,7 +30,7 @@ jobs:
           rustup default nightly
           rustup override set nightly
       - name: Clippy Check
-        uses: kristof-mattei/clippy-check@main
+        uses: actions-rs-plus/clippy-check@main
         with:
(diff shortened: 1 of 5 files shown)


.github/workflows/clippy-check.yml
```diff
@@ -21,21 +21,17 @@ jobs:
...
-          default: true
-          override: true
+        run: |
+          rustup set profile minimal
+          rustup toolchain install nightly --component clippy
+          rustup default nightly
+          rustup override set nightly
       - name: Clippy Check
-        uses: actions-rs/clippy-check@v1
+        uses: kristof-mattei/clippy-check@main
```
zedifen, line 33: Sorry, it seems that `actions-rs-plus`'s version of `clippy-check` is more actively maintained, as it's a fork by `kristof-mattei` themself ( See: https://github.com/actions-rs/clippy-check/pull/165#issuecomment-1277061858 ).

Code comparison: https://github.com/kristof-mattei/clippy-check/compare/main...actions-rs-plus:clippy-check:main

```suggestion
        uses: actions-rs-plus/clippy-check@main
```

By the way, there's a new feature which enables working directories choosing, as introduced in PR https://github.com/actions-rs/clippy-check/pull/158. This is merged by `10XGenomics` into their fork of `clippy-check`, which is forked from `actions-rs-plus`.

Issue #978, opened 2022-10-18
The Template K8s is Broken


RamboRogers:
The template K8s file uses /usr/local/bin as a bind for a volume for the V2Ray and other plugins.   Since the ss server and docker entry point exist in this folder, it is impossible for the server to start.

If you edit the file and remove the plugin volume map the system can then start.
![Screen Shot 2022-10-18 at 12 36 10 AM](https://user-images.githubusercontent.com/7763805/196336484-ff21a1b9-62a7-4f38-ae94-31fa0471d031.png)

RamboRogers:
I'll make a new one in the next few days. I made one without the plugins which works here - https://github.com/RamboRogers/shadowsocks-server-k8s

pnck:
Another solution shows at #1087

zonyitoo:
I am not very sure what the problem we are facing here, because I never use shadowsocks-rust in K8s.

PR #1087, opened 2023-01-15
[Discussing] Workaround to deploy on kubernetes with plugins

pnck:
Reference issue: #978 

SS pod cannot get started due to the mounting point at `/usr/local/bin`.  K8s' mount mechanism works different from docker's, which would overwrite the underlying files instead of keeping them togetherr. (There is no overlay filesystem.)

So the solution is either mount plugin volumes into another path, or modify the docker image to ensure that `ssserver`  at a place that won't get overwritten.

This draft shows a workaround to download and serve with plugins.

Besides the `test-connection` part is suggested removal because this pod will always be ready before the main pod, thus 
 fails invariably and making no sense.

### DIFF
--- a/k8s/shadowsocks-rust.yaml
+++ b/k8s/shadowsocks-rust.yaml
@@ -1,9 +1,17 @@
 ---
+kind: Namespace
+apiVersion: v1
+metadata:
+  name: shadowsocks-rust
+  labels:
+    app.kubernetes.io/name: shadowsocks-rust
+---
 # Source: shadowsocks-rust/templates/serviceaccount.yaml
 apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: shadowsocks-rust
+  namespace: shadowsocks-rust
   labels:
     helm.sh/chart: shadowsocks-rust-0.1.0
     app.kubernetes.io/name: shadowsocks-rust
@@ -16,6 +24,7 @@ kind: ConfigMap
 apiVersion: v1
 metadata:
   name: shadowsocks-rust
+  namespace: shadowsocks-rust
   labels:
     helm.sh/chart: shadowsocks-rust-0.1.0
     app.kubernetes.io/name: shadowsocks-rust
@@ -34,7 +43,9 @@ data:
             "password": "mypassword",
             "server": "::",
             "server_port": 8388,
-            "service_port": 80
+            "service_port": 80,
+            "plugin": "v2ray-plugin",
+            "plugin_opts": "server;host=mydomain",
           }
         ]
     }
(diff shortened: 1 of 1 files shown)


k8s/shadowsocks-rust.yaml
```diff
@@ -34,7 +43,9 @@ data:
             "password": "mypassword",
             "server": "::",
             "server_port": 8388,
-            "service_port": 80
+            "service_port": 80,
+            "plugin": "v2ray-plugin",
+            "plugin_opts": "server;host=mydomain",
```
zonyitoo, line 48: "plugin" should be removed since they were not included in the image.

k8s/shadowsocks-rust.yaml
```diff
@@ -93,6 +106,28 @@ spec:
...
+          chmod +x /download/v2ray-plugin;
+          TAG=$(wget -qO- https://api.github.com/repos/teddysun/xray-plugin/releases/latest | grep tag_name | cut -d '"' -f4);
+          wget https://github.com/teddysun/xray-plugin/releases/download/$TAG/xray-plugin-linux-amd64-$TAG.tar.gz;
+          tar -xf *.gz;
+          rm *.gz;
+          mv xray* /download/xray-plugin;
+          chmod +x /download/xray-plugin;
+        volumeMounts:
+        - name: plugins
+          mountPath: /download
```
zonyitoo, line 130: why not just make another container for these plugins?

pnck:
@zonyitoo 

let me introduce some basic conceptions. 

k8s' workload unit is called a `pod`, which may contain multiple containers but typically only one. 
since containers don't share their filesystem, multi-container pods are used for,  e.g. a website including frontend and backend servers. `fe` server and `be` server (container image) are built independently but organized into a single pod so that when at the site bootstrap, both server start synchonously.

in this ss server case instead, we want the plugins had been injected into the ssserver container before it's serving. there is no need (and should not) to make a multi-container pod, because that will isolate plugins from the main `ssserver` program. 
so the proper way is, to setup an `init container`, which prepares (download) plugins before the `ssserver` container schedules, and save the binaries in a temporary volume which belongs to the pod that can be mount to the next (the main) container. since the pod itself doesn't destroy, the temp volume is still there.

the problem we are facing is, the original specified mounting path `/usr/local/bin` contains the `ssserver` and `entrypoint.sh` executables. when the main container starts, this files will be replaced by the mounted files, because k8s isn't  providing an `overlay filesystem`, which differ from docker. and the container will fail to start because there is no entry files at all.

so here is the simplest solution that changes the mounting path to `/usr/local/sbin`, which let the `ssserver` container start normally while having plugins in another path in `$PATH`.  then the `plugin` options shall work.
 
 &nbsp;
 
but in fact, the k8s deployment is still far from available. the ss container only serves connections from inside the cluster (or on the node machine) at the moment. i'm still doing experiments on it.

zonyitoo:
Make sense.

PR #1106, opened 2023-02-04
⛔ config: deny unknown fields

database64128:


dev4u:
这个会收获一拨小白

zonyitoo:
We may have to close the “issue” tab for a month :)

database64128:
Then they'll just start flooding Discussions. 😄

Jokes aside, sing-box and my proxy projects all do this, and only one of my projects got [this one issue](https://github.com/database64128/swgp-go/issues/25), and it's most likely they used unknown fields only because the Go standard library's JSON serializer cannot deserialize JSON documents with comments, which was never an issue for shadowsocks-rust.

PR #1135, opened 2023-03-08
Update docker base image

vndroid:
- Update base image to alpine 3.17


zonyitoo:
What would changed if it was upgraded?

vndroid:
> What would be changed if it was upgraded?

Only the base image was upgraded, the user was insensitive.

The alpine 3.17 is the latest release version.

vndroid:
> > What would be changed if it was upgraded?
> 
> Only the base image was upgraded, the user was insensitive.
> 
> The alpine 3.17 is the latest release version.

Okey, by the way, updates the README.md about the build command.

Issue #1156, opened 2023-03-23
MPTCP support


skeggse:
It looks like `shadowsocks-libev` supports a `--mptcp` flag on supported Kernels, and it'd be neat to have feature parity with that flag in `shadowsocks-rust`!

zonyitoo:
Well, I didn’t realize that it was actually useful.

xinlake:
- Windows 11
- stable-x86_64-pc-windows-msvc unchanged - rustc 1.68.2 (9eb3afe9e 2023-03-27)

v1.15.2 compiles successfully and works fine, v1.15.3 compiles with errors
```
   Compiling clap v4.2.1
   Compiling xdg v2.4.1
   Compiling daemonize v0.5.0
   Compiling sysexits v0.5.0
   Compiling trust-dns-resolver v0.22.0
   Compiling shadowsocks v1.15.3 (E:\OSS\SourceCode\Client-Android-Optimize\core\src\main\rust\shadowsocks-rust\crates\shadowsocks)
error[E0425]: cannot find value `IPPROTO_MPTCP` in crate `libc`
   --> crates\shadowsocks\src\net\sys\unix\linux\mod.rs:224:64
    |
224 |         let fd = libc::socket(family, libc::SOCK_STREAM, libc::IPPROTO_MPTCP);
    |                                                                ^^^^^^^^^^^^^ help: a constant with a similar name exists: `IPPROTO_MTP`
    |
   ::: D:\Rust\.cargo\registry\src\github.com-1ecc6299db9ec823\libc-0.2.141\src\unix\linux_like\mod.rs:885:1
    |
885 | pub const IPPROTO_MTP: ::c_int = 92;
    | ------------------------------ similarly named constant `IPPROTO_MTP` defined here

For more information about this error, try `rustc --explain E0425`.
error: could not compile `shadowsocks` due to previous error
warning: build failed, waiting for other jobs to finish...
````

zonyitoo:
Why your rustc with windows target compiled the code only for Linux?

xinlake:
> Why your rustc with windows target compiled the code only for Linux?

I'm compiling Shadowsocks Rust for [Android app](https://github.com/xinlake/privch-client-flutter/tree/main/xinlake-tunnel/plugin/android) on Windows PC.
- Rust Toolchain: stable-x86_64-pc-windows-msvc
- Rust Target: aarch64-linux-android, armv7-linux-androideabi

[...]

PR #1157, opened 2023-03-24
feat(mptcp): Basic implementation of Multipath-TCP

zonyitoo:
- Currently support macOS (Client Only) and Linux (> 5.19)
  - ref #1156

### DIFF
--- a/crates/shadowsocks-service/src/config.rs
+++ b/crates/shadowsocks-service/src/config.rs
@@ -171,6 +171,7 @@ struct SSConfig {
 
     #[serde(skip_serializing_if = "Option::is_none")]
     no_delay: Option<bool>,
+
     #[serde(skip_serializing_if = "Option::is_none")]
     keep_alive: Option<u64>,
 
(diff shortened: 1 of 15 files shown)


dev4u:
请教一下，这玩意有什么用途吗？

database64128:
The MPTCP folks seem to be pushing a lot lately. MPTCP support detection logic just got merged into the Go runtime: golang/go@59d7c69342edbf8943699394513fb9b8fb175d7e

zonyitoo:
Some observation:

- Linux 5.19 (Ubuntu 22.10) doesn't support `TCP_KEEPIDLE` and `TCP_KEEPINTV` sockopts on accepted sockets. But it works for sockets after `connect`, interesting.

- macOS supports `AF_MULTIPATH` only for client sockets. `bind()` will return `EOPNOTSUPP` and `setsockopt` with `SO_REUSEADDR` will return `ENOPROTOOPT `.

```
socket(0x27, 0x1, 0x6)		 = 9 0
bind(0x9, 0x700005C74560, 0x10)		 = -1 102
```

zonyitoo:
cc https://github.com/multipath-tcp/mptcp_net-next/issues/383

wangyoucao577:
Hi, why does it require Linux > 5.19? It's MPTCPv1, which is available since Linux v5.6, right? https://github.com/multipath-tcp/mptcp_net-next/wiki#changelog

zonyitoo:
Did you tried to run it under older version of Linux Kernel?

wangyoucao577:
I haven't tried this app yet. But I have tried `mptcpize run -d iperf3` on Linux v5.15 (both client and server side), which worked well.

PR #1172, opened 2023-04-14
Make udprelay/crypto_io public

XOR-op:
Since the `ProxySocket` interface is not versatlie enough (i.e. require a ShadowUdpSocket under it), developers may need write their own structs based on cryptography tools the crate offers. However, the corresponding module `udprelay/crypto_io` is private, making this difficult. 

This pull request makes `udprelay/crypto_io` module public to solve this problem. It's also the counterpart of `tcprelay/crypto_io`, which has been public in the latest release.

### DIFF
--- a/crates/shadowsocks/src/relay/udprelay/mod.rs
+++ b/crates/shadowsocks/src/relay/udprelay/mod.rs
@@ -54,7 +54,7 @@ pub use self::proxy_socket::ProxySocket;
 mod aead;
 #[cfg(feature = "aead-cipher-2022")]
 mod aead_2022;
-mod crypto_io;
+pub mod crypto_io;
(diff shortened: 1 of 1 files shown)


zonyitoo:
How developers use the functions in `crypto_io`?

XOR-op:
For example, use `encrypt_client_payload` to encrypt udp packets.

zonyitoo:
So the connection is not UDP?

XOR-op:
I don't quite get your meaning. But developers may want other underlying "connections", such as TCP, another Shadowsocks UDP stream or some homebrew protocols. Since `ProxySocket` does not support this, the easiest way is to export the underlying cryptography methods.

PR #1194, opened 2023-05-16
Allow creating UDP sockets on iOS

pinkisemils:
I've just added support for building and running ShadowSocks on iOS. We've used it as a library, and it works well enough for our usecase. Would be great to see this upstreamed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/shadowsocks/shadowsocks-rust/1194)

### DIFF
--- a/crates/shadowsocks/src/net/udp.rs
+++ b/crates/shadowsocks/src/net/udp.rs
@@ -9,6 +9,7 @@ use std::{
     target_os = "linux",
     target_os = "android",
     target_os = "macos",
+    target_os = "ios",
     target_os = "freebsd"
 ))]
 use std::{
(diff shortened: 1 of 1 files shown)


zonyitoo:
Ok, but why? Are you guys use these functions directly without any shadowsocks protocol encapsulation?

pinkisemils (commented):
The crate just won't compile without exposing these modules for iOS targets.

> *[Reviewable](https://reviewable.io/reviews/shadowsocks/shadowsocks-rust/1194#-:-NVcu-749exPIZlnsjQG:ba2jm3q)* status: 0 of 1 files reviewed, all discussions resolved

PR #1204, opened 2023-05-27
Added configs for snap daemons

fertkir:
Currently if you run `snap start shadowsocks-rust.ssserver-daemon` it will fail because of the missing config.

This patch adds paths to config files. All a user needs to do is to add desired local/server configs to $SNAP_COMMON directory.

Still there's a problem, that only one `sslocal`/`ssserver` daemon instance is allowed.
With `ssserver` you can work around this by runing multiple servers in one process (shadowsocks config allows it).
But with `sslocal` this is, alas, impossible.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -14,19 +14,24 @@ use directories::ProjectDirs;
 use serde::Deserialize;
 
 /// Default configuration file path
-pub fn get_default_config_path() -> Option<PathBuf> {
+pub fn get_default_config_path(config_file: &str) -> Option<PathBuf> {
     // config.json in the current working directory ($PWD)
+    let config_files = vec![config_file, "config.json"];
     if let Ok(mut path) = env::current_dir() {
-        path.push("config.json");
-
-        if path.exists() {
-            return Some(path);
+        for filename in &config_files {
+            path.push(filename);
+            if path.exists() {
+                return Some(path);
+            }
+            path.pop();
         }
     } else {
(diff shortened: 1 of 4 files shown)


zonyitoo:
https://github.com/shadowsocks/shadowsocks-rust/blob/0c4f9da320e414be59aadf764f3d586ed4df59e2/src/config.rs#L59-L66

`/etc/shadowsocks-rust/config.json` is in the default path.

fertkir:
I don't think that having one config file for both `ssserver` and `sslocal` is very usable, since same config fields mean different things for server and local. AFAIU, server and local need separate configs.

zonyitoo:
Why would you run `sslocal` and `ssserver` simotanously on the same machine?

On the other hand, add these `local.json` and `server.json` into `src/config.rs`.

fertkir:
Added, please review

fertkir:
Also, do you know any way to run multiple instances of  `sslocal` snap daemon?

zonyitoo:
I don't know. I run my own services with handwriting systemd service files.

fertkir:
Me too. But when running snaps with handwritten system service, `journalctl` doesn't show the logs.

zonyitoo:
Hmm, shouldn't we also support `config.json`?

fertkir:
It is used if `server.json`/`client.json` is not provided:
``
 let config_files = vec![config_file, "config.json"];
`

PR #1223, opened 2023-06-14
added v2ray-plugin to snap

fertkir:
Now it's possible to run the following, without need to install v2ray-plugin separately:
```bash
shadowsocks-rust.ssserver --plugin v2ray-plugin --plugin-opts "server" --password "password" --server-addr 0.0.0.0:11111 --encrypt-method chacha20-ietf-poly1305
```

I don't know how to make snapcraft checkout the latest tagged commit, so I had to hard-code it.


zonyitoo:
Is this really a good way to include some binaries that doesn't belong to this project?

Are there any other ways to "mount" any binaries to snap's `PATH`?

PR #1230, opened 2023-06-25
Make clippy happier

zedifen:


zonyitoo:
Haha. But clippy still have some "suggestions". Check the "Files changed" tab.

zedifen:
@zonyitoo That's right 😂 ... Actually I did some test before submitting this PR, I apply some of clippy's suggestions, while for the rest of them, some I think are false warnings, and some I don't know whether to fix or not. 🤣

zedifen:
Oh, by the way, currently the musl cross-build would fail on stable rust toolchain with recent zig versions. Moving to rust nightly would fix it.

PR #1254, opened 2023-07-30
bugfix: acl rule pattern match error for tree rule

chuxi:
the generated acl file by running `genacl_proxy_gfw_bypass_china_ip.py` does not correctly match the pattern in `src/acl/mod.rs`.

the old version will use heavy regex set rules and it will compile all patterns when the first time used. Instead of using tree subdomain methods. 

Actually, most rules are generated for subdomain tree. By removing the last duplicated `\$`, we can make it right.

If it is not right or has some problem, please comment here.

### DIFF
--- a/crates/shadowsocks-service/src/acl/mod.rs
+++ b/crates/shadowsocks-service/src/acl/mod.rs
@@ -192,7 +192,7 @@ impl ParsingRules {
     fn add_regex_rule(&mut self, mut rule: String) {
         static TREE_SET_RULE_EQUIV: Lazy<Regex> = Lazy::new(|| {
             RegexBuilder::new(
-                r#"^(?:(?:\((?:\?:)?\^\|\\\.\)|(?:\^\.(?:\+|\*))?\\\.)((?:[\w-]+(?:\\\.)?)+)|\^((?:[\w-]+(?:\\\.)?)+))\$$"#,
+                r#"^(?:(?:\((?:\?:)?\^\|\\\.\)|(?:\^\.(?:\+|\*))?\\\.)((?:[\w-]+(?:\\\.)?)+)|\^((?:[\w-]+(?:\\\.)?)+))\$?$"#,
             )
             .unicode(false)
             .build()


zonyitoo:
You shouldn't just remove the last `\$` instead of making it optional: `\$?`

chuxi:
I also had the solution `\$?` tested. However, I noticed the generated `acl` file is using the ending `$` for regex rule set. 

To subdomain tree, it does not need ending `$`. So here I removed the `\$`. Or for some regex rule it may be confusing.

zonyitoo:
Hmm, I think it is just a special style only in https://raw.githubusercontent.com/NateScarlet/gfwlist.acl/master/gfwlist.acl.json . He omitted the tailing `$` for simple domain rules, but kept the `$` for complex rules:

```
\\.google\\.ws
^([^/]+\\.)*google\\.(ac|ad|ae|af|ai|al|am|as|at|az)$
```

It should be a special case, because if he written the first rule as `^([^/+\\.)*\\.google\\.ws$`, then it could also be handled as a tree rule.

chuxi:
I had tried the two solution: removing `\$` or updating `\$?`. 

And the source file `https://github.com/shadowsocks/shadowsocks-libev/blob/master/acl/gfwlist.acl` it has the ending `$`. `shadowsocks-rust` does not have the ending `$` for acl rules. 

So to keep compatibility, I update it again to use `\$?`.

chuxi:
> Hmm, I think it is just a special style only in https://raw.githubusercontent.com/NateScarlet/gfwlist.acl/master/gfwlist.acl.json . He omitted the tailing `$` for simple domain rules, but kept the `$` for complex rules:
> 
> ```
> \\.google\\.ws
> ^([^/]+\\.)*google\\.(ac|ad|ae|af|ai|al|am|as|at|az)$
> ```
> 
> It should be a special case, because if he written the first rule as `^([^/+\\.)*\\.google\\.ws$`, then it could also be handled as a tree rule.

Yes. I see. So I updated again.

zonyitoo:
Let me think twice..

chuxi:
> Let me think twice..

If you decided, just tell me here... 

I can accept both solutions...

Issue #302, opened 2020-10-19
Support [SIP008] Online Config


zonyitoo:
https://github.com/shadowsocks/shadowsocks-org/issues/89

This feature allows `sslocal` to update configurations dynamically from a config that hosts on a HTTPS server.

DuckSoft:
did you mean something like this?

```ss-local --sip008 '{"server":"xxxxxx","port":12345,"password":"somerandomnode","method":"aes-128-gcm"}' ....```

**Comment by zonyitoo** [comment]
```
$ sslocal --sip008 "https://URL-TO-SSCONF"
```

Maybe something like this.

fernvenue:
This is gonna be a great feature.

zonyitoo:
Because in practice, the SIP008 configuration file get from an URL usually require some modification. So what I was doing is:

1. Add a crontab script to download the SIP008 configuration file from an URL
2. Do my modifications, produce a new configuration file
3. Send a signal to `sslocal` process, and let it reload the configuration file

fernvenue:
What if make it work as an alternative of `"servers": []` part for `sslocal` and `ssservice local`? We can still do our modifications locally.

zonyitoo:
There may be quite a lot of works to do:

1. Run a background task for pulling SIP008 configurations in a fix interval. (EASY)
2. `PingBalancer` supports updating servers separately from different sources (command line, configuration file, SIP008).
3. Make HTTPS requests to SIP008 URL:
    - Which HTTP library? `reqwest` should be a choice.
    - Which TLS library? `rust-tls` for using system provided TLS framework, or `rustls`. 
    - HTTP's outbound sockets should follow the `outbound-*` configurations. (HARD)

Still wondering how to customize the sockets with `reqwest`'s API.

fernvenue:
Glad to see shadowsocks-rust supports SIP008 since [v1.16.2](https://github.com/shadowsocks/shadowsocks-rust/releases/tag/v1.16.2).

PR #1287, opened 2023-09-04
ssurl: add support for SIP008 links (ssconf links with ss contents)

k0tran:
## What is changed

ssurl now supports parsing ssconf links

## What is ssconf?

This is [outline's](https://github.com/Jigsaw-Code/outline-client) way of [providing dynamic key configuration](https://www.reddit.com/r/outlinevpn/wiki/index/dynamic_access_keys/) which is similar to [SIP008](https://github.com/shadowsocks/shadowsocks-rust/issues/302).

The `ssconf://` link is the same as `https://` link to `ss://` configuration (`json` in this implementation is not supported though it is easy to implement in the future)

## Motivation

Outline is based on shadowsocks protocol but has no official cli interface (for clients). My provider is using dynamic keys and since [shadowsocks-rust#302](https://github.com/shadowsocks/shadowsocks-rust/issues/302) (SIP008) is WIP this "hack" will allow using shadowsocks-rust with dynamic keys at least for now (generating new config easily once in a while).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -675,6 +675,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "encoding_rs"
+version = "0.8.33"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7268b386296a025e474d5140678f75d6de9493ae55a5d709eeb9dd08149945e1"
+dependencies = [
+ "cfg-if",
+]
+
+[[package]]
 name = "enum-as-inner"
 version = "0.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


zonyitoo:
I don't think it should be mixed into the standard SIP002 URL.

You could make a separated function for parsing SIP008 styled URLs.

k0tran:
I've got notified about reply by email, but comment seems to be deleted now. I think it's mentioned important thing which should be discussed.

The reply stated that it's a bad decision to mix two URL formats together (SIP002 and SIP008). Therefore I see two ways of fixing this problem in current implementation:

1. Use a separate flag for specifying `ssconf` links (for example `-r` and `--remote`)
    - The good thing is that now `-d` guarantees no GET requests
    - On the other hand we will have two flags for decoding urls which prevents universal usage. It's inconsistent and links already differ by prefix (`ss` and `ssconf`) why separate then?
2. Use a special flag **with `-d`** to turn off/on remote links (for example `--no-remote`)
    - This way we can allow to prevent potential risks of confusing `ss` and `ssconf` while also providing consistent interface
    - We can also add `--allow-remote` flag and don't accept `ssconf` links without it, but since links already differ I'm leaning more towards `--no-remote` implementation.

I will implement the solution after it's decided.

zonyitoo:
I personally prefer 1. Because `ssconf://` is just one special case of URL specifically for outline. Maybe you can just name the command line argument as "--outline-remote".

zonyitoo:
`reqwest` will compile a full functional SSL library into the `ssurl` binary. It should be enabled by an optional feature `--features "utility-url-ssconf`.

PR #1309, opened 2023-09-19
fix build on ios

ibigbug:


ibigbug:
nvm, it seems to have been fixed already.

ibigbug:
yeah. haven't synced it up since a while, ty for the prompt anyways

Issue #1330, opened 2023-10-19
feature request: support macOS's launch_activate_socket


liudongmiao:
In shadowsocks-libev, there is `launch_activate_socket` in https://github.com/shadowsocks/shadowsocks-libev/pull/1003.
In shadowsocks-rust, there is no such feature.

So, it's possible for shadowsocks-rust to support this feature?

There is another rust project https://github.com/LinkTed/doh-client ships with `launch_activate_socket`.

zonyitoo:
It seems that you only create the TCP socket via `launch_activate_socket`, how about the UDP socket?

zonyitoo:
https://github.com/LinkTed/doh-client/blob/master/src/listen/activation_socket.rs#L32

It has memory leaks if `launch_activate_socket` returns `cnt > 1`.

liudongmiao:
@zonyitoo I will try to implement and make a PR.

PR #1331, opened 2023-10-20
support launch_activate_socket on macos

liudongmiao:
In macOS 10.10+, `launch_activate_socket` was introduced.
So whenever the port is requested, the `launchd` can launch the program dynamically. A possible plist may like this:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>Debug</key>
	<true/>
	<key>EnablePressuredExit</key>
	<true/>
	<key>Label</key>
	<string>com.github.shadowsocks-rust</string>
        <key>ProgramArguments</key>
        <array>
                <string>/path/to/sslocal</string>
                <!-- -b can be ommited, as launchd takes over it -->
                <!-- other arguments -->
		<string>-c</string>
		<string>/path/to/config</string>
        </array>
        <key>Sockets</key>
        <dict>
                <!-- the name is hard coded in sslocal -->
                <key>Listeners</key>
                <dict>
                        <key>SockNodeName</key>
                        <string>0.0.0.0</string> <!-- or 127.0.0.1 -->
                        <key>SockServiceName</key>
                        <string>socks</string> <!-- port name (/etc/services) or port number -->
			<key>SockType</key>
			<string>stream</string> <!-- currently, only tcp is supported -->
                </dict>
        </dict>
	<key>StandardErrorPath</key>
	<string>/path/to/stderr</string>
	<key>StandardOutPath</key>
	<string>/path/to/stdout</string>
</dict>
</plist>
```

The magic is `launchd` will listen to the `SockServiceName`, when the port is requested, the `sslocal` will be launched.

### DIFF
--- a/crates/shadowsocks/src/net/sys/unix/bsd/macos.rs
+++ b/crates/shadowsocks/src/net/sys/unix/bsd/macos.rs
@@ -227,15 +227,29 @@ extern "C" {
     fn launch_activate_socket(name: *const libc::c_char, fds: *mut *mut libc::c_int, cnt: *mut libc::size_t) -> libc::c_int;
 }
 
-pub fn get_launchd_socket() -> io::Result<RawFd> {
+fn get_launchd_socket_fd(socktype: libc::c_int) -> io::Result<RawFd> {
     let name = std::ffi::CString::new("Listeners").expect("Listeners");
     let mut fds = ptr::null_mut();
     let mut count = 0;
-    let error =  unsafe { launch_activate_socket(name.as_ptr(), &mut fds, &mut count) };
+    let error = unsafe { launch_activate_socket(name.as_ptr(), &mut fds, &mut count) };
     match error {
         0 => {
             let result = match count {
-                1 => Ok(unsafe { *fds.offset(0) }),
+                1 => {
+                    let socket_fd = unsafe { *fds.offset(0) };
+                    let mut option_value: libc::c_int = 0;
+                    let mut option_len = mem::size_of::<libc::c_int>() as libc::socklen_t;
+                    unsafe { libc::getsockopt(socket_fd, libc::SOL_SOCKET, libc::SO_TYPE, &mut option_value as *mut _ as *mut _, &mut option_len) };
+                    if option_value == socktype {
+                        let mut nonblocking = true as libc::c_int;
+                        unsafe { libc::ioctl(socket_fd, libc::FIONBIO, &mut nonblocking) };
+                        Ok(socket_fd)
+                    } else {
+                        let msg = format!("launch_activate_socket, unexpected socket type: {} (expect {})", option_value, socktype);
+                        warn!("{}", msg);
+                        Err(io::Error::new(ErrorKind::Other, msg))
+                    }
+                }
                 _ => {
                     let msg = format!("launch_activate_socket, unexpected sockets: {}", count);
                     warn!("{}", msg);
@@ -249,13 +263,13 @@ pub fn get_launchd_socket() -> io::Result<RawFd> {
             match error {
                 libc::ENOENT => {
                     warn!("launch_activate_socket: The socket name specified does not exist in the caller's launchd.plist(5).")
-                },
+                }
                 libc::ESRCH => {
                     // debug!("launch_activate_socket: The calling process is not managed by launchd(8).")
-                },
+                }
                 libc::EALREADY => {
                     warn!("launch_activate_socket: The specified socket has already been activated.")
-                },
+                }
                 _ => {
                     warn!("launch_activate_socket: {}", io::Error::from_raw_os_error(error))
                 }
@@ -267,11 +281,9 @@ pub fn get_launchd_socket() -> io::Result<RawFd> {
 
 /// Create a TCP socket for listening
 pub async fn create_inbound_tcp_socket(bind_addr: &SocketAddr, _accept_opts: &AcceptOpts) -> io::Result<TcpSocket> {
-    if let Ok(launchd_socket) = get_launchd_socket() {
-        log::info!("launch_activate_socket, use launched socket for {}", bind_addr);
-        let mut nonblocking = true as libc::c_int;
-        unsafe { libc::ioctl(launchd_socket, libc::FIONBIO, &mut nonblocking) };
-        return Ok(unsafe { TcpSocket::from_raw_fd(launchd_socket) });
+    if let Ok(launchd_socket_fd) = get_launchd_socket_fd(libc::SOCK_STREAM) {
+        log::info!("create_inbound_tcp_socket, use launched socket for {}", bind_addr);
+        return Ok(unsafe { TcpSocket::from_raw_fd(launchd_socket_fd) });
     }
     match bind_addr {
         SocketAddr::V4(..) => TcpSocket::new_v4(),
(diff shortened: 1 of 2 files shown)


crates/shadowsocks/src/net/tcp.rs
```diff
@@ -145,12 +145,27 @@ impl TcpListener {
...
+                _ => {
+                    log::info!("create_inbound_tcp_socket, addr: {}, bound: {}", addr, local_addr);
+                    true
+                },
+            },
+            Err(_) => false,
+        };
+        #[cfg(not(target_os = "macos"))]
+        let bound = false;
+        if !bound {
```
liudongmiao, line 161: any better way to prevent bind again for `launch_activate_socket`?

crates/shadowsocks/src/net/sys/unix/bsd/macos.rs
```diff
@@ -223,8 +223,56 @@ pub fn set_tcp_fastopen<S: AsRawFd>(socket: &S) -> io::Result<()> {
...
+            unsafe { libc::free(fds as *mut _) };
+            result
+        }
+        error => {
+            match error {
+                libc::ENOENT => {
+                    warn!("launch_activate_socket: The socket name specified does not exist in the caller's launchd.plist(5).")
+                },
+                libc::ESRCH => {
+                    // debug!("launch_activate_socket: The calling process is not managed by launchd(8).")
```
liudongmiao, line 254: should we add log if not managed via launchd?

liudongmiao (commented):
question for pr

liudongmiao:
This should close #1330

liudongmiao:
This PR doesn't omit the port in the arguments, so, "-b 0.0.0.0:0" or any similiar argument should be passed, however, it won't be used.

zonyitoo:
Hmm.. I have also did some works about `launch_active_socket` in my local branch. Let me finish and invite you to review.

zonyitoo:
Prefer: https://github.com/shadowsocks/shadowsocks-rust/pull/1332

liudongmiao:
@zonyitoo why #1332 is so complex...

In this pr, only tcp is supported, however, I don't think udp should be supported.
And, I only test with socks and http protocol.

zonyitoo:
Because

1. the `name` of the socket should be configurable.
2. all the `protocol`s may have different implementation details.

> In this pr, only tcp is supported, however, I don't think udp should be supported.

Why UDP shouldn't be supported?

liudongmiao:
> Why UDP shouldn't be supported?

UDP should be simple too. However, we should consider for the real use case.

Why `launch_activate_socket` for `sslocal`?
IMO, it's used for start up automatically in macOS.

For example, browser use socks or http proxy, and whenever it want to use the proxy, it try to connect the port, launchd would start the process.

There are several use case for sslocal:
- `socks`, use as socks proxy
   - socks is a `tcp` proxy that supports `udp`
- `local-http`, use as http proxy, no `udp`
- `local-tunnel`, what's the use case?
- `local-redir`, what's the use case?
- `local-dns`, use as dns server?
   - As I know, dns server is on udp default, why tcp?
- `local-tun`, don't use tcp / udp at all.

So, when this feature is implemented, what's the use case?
`sslocal` supports multi local configurations, why macOS need this, and implement in a complex configuration like #1332?

zonyitoo:
> `local-tunnel`, what's the use case?

Tunnel is for establishing a pure shadowsocks end-to-end tunnel.

> `local-redir`, what's the use case?

`sslocal` starts when user set the ipfw/... firewall configuration.

> `local-dns, use as dns server?`

DNS server requires both TCP and UDP.

liudongmiao:
@zonyitoo Just support udp. Use `getsockopt` to check whether fd is `tcp` or `udp`.
// Sorry, I didn't test udp mode yet.

However, my implementation is still `Listeners`, no `TCPListener` nor `UDPListener`.
Actually, `launchd.plist` even supports unix domain, should there be `UNIXListener`?
A single `Listeners`or `Listener` should be easy for macOS users to configure, as you say, there should be only one activate socket in launchd.
// I use `Listeners`, as it's the way implemented in `shadowsocks-libev`, however, I don't think there are many users know about this.

So, when macOS users want to use launchd, add a launchd.plist.
And if they don't want to use launchd.plist, they can use the same arguments with launchd.plist.

zonyitoo:
> Actually, `launchd.plist` even supports unix domain, should there be `UNIXListener`?

I don't think so. There is no local instance supports unix domain socket listener.

> A single `Listeners` or `Listener` should be easy for macOS users to configure, as you say, there should be only one activate socket in launchd.

I just tested on my laptop: launchd supports multiple activate sockets. `sslocal` will start when any of them were accessed. `launch_activate_socket` was able to return all the activate sockets even they are not accessed.

Further more, it is Ok for you to use `Listeners` as the socket name:

```json
{
    "locals": {
        // These are required because SOCKS5 UDP Associate command has to return the IP:Port to clients
        "local_address": "127.0.0.1",
        "local_port": 1080,
        "launchd_tcp_socket_name": "Listeners"
    }
}
```

It works exactly the same.

zonyitoo:
Since #1332 have included the feature in this PR, so I am going to close it and merge #1332 . Further discussion moved to #1330 .

liudongmiao:
@zonyitoo it seems doesn't work for me.

plist
```plist
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
        <key>Debug</key>
        <true/>
        <key>EnablePressuredExit</key>
        <true/>
        <key>Label</key>
        <string>com.github.shadowsocks-3128</string>
        <!--
        <key>EnvironmentVariables</key>
        <dict>
            <key>RUST_BACKTRACE</key>
            <string>full</string>
        </dict>
        -->
        <key>ProgramArguments</key>
        <array>
                <string>/usr/local/bin/sslocal</string>
                <string>-b</string>
                <string>0.0.0.0:0</string> <!-- if i don't specify, sslocal-rust will be exit -->
                <string>--protocol</string>
                <string>http</string>
                <string>-c</string>
                <string>/usr/local/etc/shadowsocks-libev.json</string>
        </array>
        <key>Sockets</key>
        <dict>
                <key>Listeners</key>
                <dict>
                        <key>SockNodeName</key>
                        <string>0.0.0.0</string>
                        <key>SockServiceName</key>
                        <string>3128</string>
                        <key>SockType</key>
                        <string>stream</string>
                </dict>
        </dict>
        <key>StandardErrorPath</key>
        <string>/tmp/ss-rust.log</string>
        <key>StandardOutPath</key>
        <string>/tmp/ss-rust.log</string>
</dict>
</plist>
```

and the configuration, as I have only a local.

```
{
  "launchd_tcp_socket_name": "Listeners",
  "method": "none",
  "plugin": "xxx",
  "plugin_opts": "xxx",
  "server": "xxx",
  "server_port": 443,
  "timeout": 60
}
```

BTW, my modification works well. And actually, I'm not familiar with the multi configuration of shadowsocks-rust.

zonyitoo:
```json
{
    "locals": [
        {
            "launchd_tcp_socket_name": "Listeners", ...
        }
    ]
}
```

liudongmiao:
@zonyitoo Thanks. However, it doesn't response at all.
Then I found it doesn't set nonblocking for the fd, I have make a in #1451.

PR #1350, opened 2023-11-11
Parse tun_interface_destination from ssconfig

AzazKamaz:


zonyitoo:
What is the actual use case in real life?

PR #1451, opened 2024-03-05
set nonblocking for launch_activate_socket

liudongmiao:
In https://github.com/shadowsocks/shadowsocks-rust/issues/1330 , shadowsocks-rust introduces `launch_activate_socket` for the request of https://github.com/shadowsocks/shadowsocks-rust/issues/1330 .
However, #1330 seems forget to set nonblocking as similiar pr in https://github.com/shadowsocks/shadowsocks-rust/pull/1331 .

### DIFF
--- a/crates/shadowsocks-service/src/sys/unix/macos.rs
+++ b/crates/shadowsocks-service/src/sys/unix/macos.rs
@@ -74,6 +74,8 @@ pub fn get_launch_activate_socket(name: &str) -> io::Result<RawFd> {
     } else {
         // Take fds[0] as the result
         let fd = unsafe { *fds };
+        let mut nonblocking = true as libc::c_int;
+        unsafe { libc::ioctl(fd, libc::FIONBIO, &mut nonblocking) };
         Ok(fd as RawFd)
     };
 


zonyitoo:
https://github.com/shadowsocks/shadowsocks-rust/blob/b5c74a6b1de380c2cdb26cee69a58a5b15029234/crates/shadowsocks-service/src/local/tunnel/tcprelay.rs#L55-L56

The key issue is tokio's `TcpListener::from_std` doesn't set nonblock. The `get_launch_activate_socket` is for getting an TcpListener from libstd, which doesn't need nonblock.

I am wondering if there is another better way.

liudongmiao:
Yes, when I try to support, I try to find way to set nonblock...
And I found it's hard to set non block with fd in `Tokio` and internal library (Sorry, I have forgotten the name).

Then, I just make the fd nonblock use unix way. The `FIONBIO` is refered by rust-lang.
// There are two way to set nonblock in rust, FIONBIO and O_NONBLOCK
// fd.rs use O_NONBLOCK, net.rs use FIONBIO.

liudongmiao:
Thanks!

PR #1468, opened 2024-03-21
resolve #1467: build sswinservice for Windows by default

### LINKED ISSUE #1467: Default build `sswinservice` for Windows releases
The binary to register as a Windows service added in #1292. However, it's not default built. It would be great if you can set it as default features for release target, so that users can take advantage of GitHub CI.

amorphobia:


zonyitoo:
Emm, why not also add the feature for `windows-gnu` builds?

amorphobia:
> Emm, why not also add the feature for `windows-gnu` builds?

Sure. I'll also add for it. Please allow me some time.

PR #1491, opened 2024-04-06
Fix per-server outbound options not taking effect (part 2)

stormynoct:
I'm in a multi-WAN environment and I want the two WAN connections to backup each other. I tried to set outbound_fwmark for each server but it seems not working as expected.

By looking into the code I found two issues:
1. PingBalancer does not set the fwmark correctly in connect_opts so it's not able to detect whether the second WAN is working or not.
2. The outbound traffic seems to ignore the fwmark options in some cases. The connect_opts are modified correctly in some classes while not in others.

I drafted this change containing the following fixes:
1. Add outbound_bind_addr and outbound_bind_interface into ServerInstanceConfig
2. Refactor PingBalancer to hold a copy of per-server ConnectOpts
3. Add new interfaces for specifying ConnectOpts in AutoProxyClientStream

It works for my case, but I don't have a good environment to be able to test all the logic I added. Also the related code pieces are quite a lot, so not sure if I missed any place.

### DIFF
--- a/crates/shadowsocks-service/src/config.rs
+++ b/crates/shadowsocks-service/src/config.rs
@@ -1888,7 +1888,7 @@ impl Config {
                     acl: None,
                     #[cfg(any(target_os = "linux", target_os = "android"))]
                     outbound_fwmark: config.outbound_fwmark,
-                    outbound_bind_addr: outbound_bind_addr,
+                    outbound_bind_addr,
                     outbound_bind_interface: config.outbound_bind_interface.clone(),
                 };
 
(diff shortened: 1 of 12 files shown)


zonyitoo:
This is quite a big PR...

stormynoct:
Though 1 is needed in my environment, I feel like it's quite isolated from 2/3. If you think the PR is too large to review, maybe I can consider to break it down into two. One containing 2/3 and another on top containing only 1.

Maybe we can discuss first on whether you think the change 2/3 makes sense. There is some historical issues that fwmark is placed into Config while server params like addr/pass/method are placed into ServerConfig. ServerInstanceConfig looks like a bridge between these two and can move some additional parameters in Config to be per-server params.

But there is a disconnection that most places initiating an outgoing traffic uses the default ConnectOpts, they does not read ServerInstanceConfig which contains the correct information for outbound traffic. That's why I propose change 2. Both ServerInstanceConfig and cached ConnectOpts should retain a copy in PingBalancer.

zonyitoo:
So your idea is to add a `ConnectOpts` in `ServerConfig` right? LGTM. Please resolve the conflicts and I am Ok with the other changes.

Issue #1543, opened 2024-05-29
Transparently proxying UDP traffic with pf?


Orum:
This is a similar symptom to my earlier issue (#1473), but as the circumstances have drastically changed I thought I'd open a new issue.  To summarize the changes, I am no longer running `sslocal` on the host (Linux/iptables) machine; it's instead being run on my FreeBSD router which uses pf.

Obviously, this requires different firewall rules to work, but there is little if any documentation I could find for transparently proxying via pf within the shadowsocks-rust project.  What I've done here is largely guesswork, but this is the rule I came up with to transparently proxy both TCP & UDP to `sslocal` (and then on to `ssserver`):
`rdr on $int_if inet proto { tcp, udp } to !<private> -> 127.0.0.1 port 1080`

`sslocal` is being run on the router via the following command:
`sslocal -b 127.0.0.1:1080 -U --protocol redir -s 192.168.x.y:8388 -m none --tcp-redir pf --udp-redir pf`

...and finally, `ssserver` is run with `-U -m none -b  192.168.x.y:8388`

This "works" in the sense that both TCP & UDP traffic are being redirected to `sslocal` (unlike in #1473 where only TCP traffic was ever reaching `sslocal`), which is then sending the traffic on to `sserver`.  However, once again, only TCP traffic is being fully proxied correctly.

UDP traffic does arrive at `sslocal` without any doubt, as it shows up when running `sslocal` with `-vvv` like so (timestamps removed for brevity, and the host system's address being substituted here with simply `<host>`):
```
TRACE  tokio-runtime-worker ThreadId(05) shadowsocks::relay::udprelay::proxy_socket: crates/shadowsocks/src/relay/udprelay/proxy_socket.rs:235: UDP server client send to 127.0.0.1:1080, control: UdpSocketControlData { client_session_id: 7043831170210357480, server_session_id: 0, packet_id: 1, user: None }, payload length 96 bytes, packet length 103 bytes
TRACE  tokio-runtime-worker ThreadId(05) shadowsocks_service::local::redir::udprelay: crates/shadowsocks-service/src/local/redir/udprelay/mod.rs:307: received UDP packet from <host>:44849, destination 127.0.0.1:1080, length 96 bytes
[...]

PR #1567, opened 2024-06-24
add support for OpenBSD/pf

ge9:
Added support for OpenBSD/pf.
In OpenBSD's pf, `divert-to` is used for transparent proxying.
As described in https://man.openbsd.org/pf.conf, TCP destination address can be obtained by getsockname() (like Linux+TPROXY or FreeBSD+ipfw), and UDP destination address can be obtained by special socket options like IP_RECVDSTADDR.
`send_to` using "fake" source address fails, but using "0.0.0.0:0" instead works. Note that this is not related to IPv4-mapped IPv6 address issue (https://github.com/shadowsocks/shadowsocks-rust/issues/1543), because OpenBSD doesn't support IPv4-mapped IPv6 address (https://man.openbsd.org/inet6) and shadowsocks can detect it.
Currently IPv6 is not supported. Also, ` --features "local-tun"` doesn't work.


crates/shadowsocks-service/src/local/net/udp/association.rs
```diff
@@ -625,6 +625,8 @@ where
                 if bypassed { "bypassed" } else { "proxied" },
                 err
             );
+            #[cfg(target_os = "openbsd")]
+            let _ = self.respond_writer.send_to(self.peer_addr, &Address::SocketAddress(SocketAddr::new(std::net::IpAddr::V4(std::net::Ipv4Addr::new(0,0,0,0)), 0)), data).await;
```
zonyitoo, line 629: Why OpenBSD needs this "faked" destination address? `addr` is the actual address of client.

crates/shadowsocks-service/src/local/redir/sys/unix/mod.rs
```diff
@@ -15,7 +15,7 @@ cfg_if! {
         #[allow(dead_code, non_upper_case_globals, non_snake_case, non_camel_case_types)]
         #[allow(clippy::useless_transmute, clippy::too_many_arguments, clippy::unnecessary_cast)]
         mod pfvar;
-    } else if #[cfg(target_os = "freebsd")] {
+    } else if #[cfg(any(target_os = "freebsd", target_os = "openbsd"))] {
         #[path = "pfvar_bindgen_freebsd.rs"]
```
zonyitoo, line 19: `pfvar_bindgen_freebsd.rs` is generated on FreeBSD. I suggest you to generate one on OpenBSD with [bindgen-cli](https://crates.io/crates/bindgen-cli).

zonyitoo:
What is that `ktrace.out` file?

crates/shadowsocks-service/src/local/net/udp/association.rs
ge9: I think `peer_addr` is client (local) address and `addr` is remote address, which is not assigned to sslocal itself, resulting an error (Can't assign requested address, error 49).

ge9:
Ah sorry that's a stack trace...
I'll review and resend this PR sorry

ge9:
And I may be able to add IPv6 support soon.

crates/shadowsocks-service/src/local/net/udp/association.rs
zonyitoo: That's not true. `send_received_respond_packet` is for sending back packets received from remote server to local client. So:

1. `peer_addr` is the receiver (client)'s address
2. `addr` is the sender (remote server)'s address

The code in `respond_write.send_to`, which is `redir/udprelay/mod.rs:UdpRedirInboundWriter::send_to` were doing:

1. Create an UDP socket, `bind()` to `addr`
2. Call `sendto` to `peer_addr`

This is the way to make an UDP packet that source address is `addr` and destination address is `peer_addr`.

If socket's `bind()` failed right here (`EADDRNOTAVAIL`), `SO_BINDANY` is not set or working as expected.

crates/shadowsocks-service/src/local/redir/sys/unix/mod.rs
ge9: Ah actually, for bindany, I just reversed the comment-out of the old existing code. Maybe that's not working correctly? I'll look into it.

zonyitoo: WRONG COMMENT. :)

ge9: Sorry I'm not familiar with this tool, but how can I do this? Is there a commandline for generating this?

crates/shadowsocks-service/src/local/net/udp/association.rs
ge9: Sorry, I mistakenly disable set_bindany() function. If I enabled it, the original `send_to` worked correctly.
https://man.openbsd.org/getsockopt.2 says that we need `divert-reply` along with SO_BINDANY, but in my environment, it seems working correctly without `divert-reply`...🤔

crates/shadowsocks-service/src/local/redir/sys/unix/mod.rs
zonyitoo: Is `pfvar.h` in OpenBSD included in release? If no, you should first download it from its source.

Create a header file, that only has `#include <pfvar.h>`.

```bash
$ cargo install bindgen-cli
$ bindgen <your_created_header_file> -o pfvar_bindgen_openbsd.rs
```

More reference: https://rust-lang.github.io/rust-bindgen/command-line-usage.html

crates/shadowsocks-service/src/local/net/udp/association.rs
zonyitoo: `SO_BINDANY` allows socket to `bind()` an address that is not belongs to local environment. `pf` should be one of the use cases.

zonyitoo:
You should install an `rustfmt`, with:

```
rustup component add rustfmt
rustup component add rustfmt --toolchain nightly
```

Run `rustfmt` with:

```
cargo +nightly fmt
```

This project have set some unstable options, which requires nightly Rust.

ge9:
```
error[E0425]: cannot find value `IPV6_RECVDSTPORT` in crate `libc`
   --> crates/shadowsocks-service/src/local/redir/udprelay/sys/unix/bsd.rs:272:37
    |
272 |         libc::IPPROTO_IPV6 => libc::IPV6_RECVDSTPORT,
    |                                     ^^^^^^^^^^^^^^^^ not found in `libc`
    |
note: constant `crate::local::redir::sys::unix::pfvar::IPV6_RECVDSTPORT` exists but is inaccessible
   --> crates/shadowsocks-service/src/local/redir/sys/unix/pfvar_bindgen_openbsd.rs:589:1
    |
589 | pub const IPV6_RECVDSTPORT: u32 = 64;
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not accessible

```
Sorry, I can't figure out how to solve this. Do you have any idea?

I think I successfully created `pfvar_bindgen_openbsd` and specified it in mod.rs
```
cfg_if! {
    if #[cfg(any(target_os = "macos",
                 target_os = "ios"))] {
        #[path = "pfvar_bindgen_macos.rs"]
        #[allow(dead_code, non_upper_case_globals, non_snake_case, non_camel_case_types)]
        #[allow(clippy::useless_transmute, clippy::too_many_arguments, clippy::unnecessary_cast)]
        mod pfvar;
    } else if #[cfg(target_os = "freebsd")] {
        #[path = "pfvar_bindgen_freebsd.rs"]
        #[allow(dead_code, non_upper_case_globals, non_snake_case, non_camel_case_types)]
        #[allow(clippy::useless_transmute, clippy::too_many_arguments, clippy::unnecessary_cast)]
        mod pfvar;
    } else if #[cfg(target_os = "openbsd")] {
        #[path = "pfvar_bindgen_openbsd.rs"]
        #[allow(dead_code, non_upper_case_globals, non_snake_case, non_camel_case_types)]
        #[allow(clippy::useless_transmute, clippy::too_many_arguments, clippy::unnecessary_cast)]
        mod pfvar;
    }
}

```
but it seems not working...

ge9:
There are already some hard-coded constant values for FreeBSD, like https://github.com/shadowsocks/shadowsocks-rust/blob/6d7bc358c11701c7114d877cb0ab2c9e33146db9/crates/shadowsocks-service/src/local/redir/udprelay/sys/unix/macos.rs#L167. So, maybe pfvar_bindgen_freebsd.rs is not actually used either.
Should we write `pub mod pfvar;` instead of `mod pfvar` and use the constants in udprelay/sys/unix/bsd.rs?

zonyitoo:
The first error indicates that OpenBSD doesn't have `libc::IPV6_RECVDSTPORT` in [libc](https://github.com/rust-lang/libc). Is that true? OpenBSD doesn't have `IPV6_RECVDSTPORT`?

If so, IPv6 support on OpenBSD have to be disabled.

Those hard-coded constant values exists because they are not defined in [libc](https://github.com/rust-lang/libc), so I copied them directly from source.

I would suggest to add `IP_DONTFRAG` for FreeBSD, OpenBSD independently instead of using the generated pfvar.rs, because they have no relation.

ge9:
Sorry, I pasted only the part of errors, but similar error happens for libc::IP_RECVDSTPORT. 
These constants do exist in OpenBSD's C library, in `netinet/in.h`, and so included in generated pfvar_xxx.rs, but they seem not defined in Rust's [libc](https://github.com/rust-lang/libc).
The situation is exactly the same for IP_DONTFRAG. It's defined in FreeBSD`s `netinet/in.h` and included in pfvar_xxx.rs, but not defined in Rust's libc.

zonyitoo:
It would be great to open an PR to Rust's libc crate and add all those contants.

Before it was getting merge, it is Ok to just hard-coded.

ge9:
Also, I added support for IPv6, but strangely enough, transparent proxying only works **every other time**. (I don't know detailed conditions)
I think this is OpenBSD's issue, because shadowsocks with -vvvv doesn't detect any packet.

crates/shadowsocks-service/src/local/redir/udprelay/sys/unix/bsd.rs
```diff
@@ -337,6 +386,66 @@ fn get_destination_addr(msg: &libc::msghdr) -> io::Result<SocketAddr> {
...
+        let (_, addr) = SockAddr::try_init(|dst_addr, dst_addr_len| {
+            let mut cmsg: *mut libc::cmsghdr = libc::CMSG_FIRSTHDR(msg);
+            while !cmsg.is_null() {
+                let rcmsg = &*cmsg;
+                match (rcmsg.cmsg_level, rcmsg.cmsg_type) {
+                    (libc::IPPROTO_IP, libc::IP_RECVDSTADDR) => {
+                        let toaddr_in = &mut *(dst_addr as *mut libc::sockaddr_in);
+                        ptr::copy_nonoverlapping(
+                            libc::CMSG_DATA(cmsg),
+                            &(*toaddr_in).sin_addr as *const _ as *mut _,
```
zonyitoo, line 401: You should use:

```rust
use std::ptr::addr_of_mut;

                        ptr::copy_nonoverlapping(
                            libc::CMSG_DATA(cmsg),
                            addr_of_mut!((*toaddr_in).sin_addr),
```

crates/shadowsocks-service/src/local/redir/udprelay/sys/unix/bsd.rs
```diff
@@ -288,11 +335,13 @@ fn set_socket_before_bind(addr: &SocketAddr, socket: &Socket) -> io::Result<()>
     set_ip_origdstaddr(level, socket)?;
 
     // 3. disable IP fragmentation
+    #[cfg(not(target_os = "openbsd"))]//seems unsupported in OpenBSD
```
zonyitoo, line 338: Only FreeBSD-like system supports IP_DONTFRAG, which are:

```rust
#[cfg(any(target_os = "freebsd",
                  target_os = "dragonfly",
                  target_os = "macos",
                  target_os = "ios",
                  target_os = "watchos",
                  target_os = "tvos"))]
```

zonyitoo: Maybe we should separate them into 3 files:

1. macos.rs
2. freebsd.rs (freebsd-like)
3. openbsd.rs (netbsd-like)

crates/shadowsocks-service/src/local/redir/udprelay/sys/unix/bsd.rs
```diff
@@ -337,6 +386,66 @@ fn get_destination_addr(msg: &libc::msghdr) -> io::Result<SocketAddr> {
...
+                    }
+                    (libc::IPPROTO_IPV6, 64) => {//TODO: use libc::IPV6_RECVDSTPORT
+                        let toaddr_in = &mut *(dst_addr as *mut libc::sockaddr_in6);
+                        ptr::copy_nonoverlapping(
+                            libc::CMSG_DATA(cmsg),
+                            &(*toaddr_in).sin6_port as *const _ as *mut _,
+                            mem::size_of::<libc::in_port_t>(),
+                        );
+                        //assuming address comes first, and then port
+                        return Ok(());
```
zonyitoo, line 435: Your assumption is correct: https://github.com/openbsd/src/blob/3d310523b415eeee9db46a5b67eecf8f9fdd5c8f/sys/netinet/udp_usrreq.c#L662-L687

`RECVDSTPORT` comes after `RECVDESTADDR`, but we should not rely on that, I suppose.

Just let the loop run until the end.

crates/shadowsocks-service/src/local/redir/udprelay/sys/unix/bsd.rs
```diff
@@ -337,6 +386,66 @@ fn get_destination_addr(msg: &libc::msghdr) -> io::Result<SocketAddr> {
...
+                        let toaddr_in = &mut *(dst_addr as *mut libc::sockaddr_in);
+                        ptr::copy_nonoverlapping(
+                            libc::CMSG_DATA(cmsg),
+                            &(*toaddr_in).sin_addr as *const _ as *mut _,
+                            mem::size_of::<libc::in_addr>(),
+                        );
+                        toaddr_in.sin_family = libc::AF_INET as u8;
+                        *dst_addr_len = mem::size_of::<libc::sockaddr_in>() as libc::socklen_t;
+                    }
+                    (libc::IPPROTO_IP, 33) => {//TODO: use libc::IP_RECVDSTPORT
```
zonyitoo, line 407: These constants could just put in the global scope of this file.

crates/shadowsocks-service/src/local/redir/udprelay/sys/unix/mod.rs
```diff
@@ -7,7 +7,7 @@ cfg_if! {
     } else if #[cfg(target_os = "macos")] {
         mod macos;
         pub use self::macos::*;
-    } else if #[cfg(any(target_os = "freebsd"))] {
+    } else if #[cfg(any(target_os = "freebsd", target_os = "openbsd"))] {
```
zonyitoo, line 10: As I mentioned above, we should separate them into 2 different files.

crates/shadowsocks-service/src/local/tun/mod.rs
```diff
@@ -27,7 +27,8 @@ cfg_if! {
                  target_os = "linux",
                  target_os = "android",
                  target_os = "windows",
-                 target_os = "freebsd"))] {
+                 target_os = "freebsd",
+                 target_os = "openbsd"))] {
```
zonyitoo, line 31: `tun2` doesn't support openbsd: https://github.com/tun2proxy/rust-tun/blob/5b6a80a859da55a65b323f6dbfcf2893744caaec/src/platform/mod.rs#L20-L51

You could open an PR to `tun2`, they are actively responding.

crates/shadowsocks-service/src/local/redir/udprelay/sys/unix/bsd.rs
ge9: If we let the loop run, "missing destination address in msghdr" error will be raised?

zonyitoo: Well, currently yes. Maybe you can add two `bool` flag to indicates that we have already obtained addr and port.

zonyitoo:
Consider update your branch with the latest master's commit.

crates/shadowsocks-service/src/local/redir/udprelay/sys/unix/bsd.rs
ge9: That's an idea, but some functions are used from both OSes. Can I copy&paste them?

ge9:
Now I'm working on complete this pull request.
Should I separate bsd.rs into openbsd.rs and freebsd.rs as you suggested?

zonyitoo:
> Now I'm working on completing this pull request. Should I separate bsd.rs into openbsd.rs and freebsd.rs as you suggested?

I would prefer the separation. openbsd and freebsd has too many differences.

ge9:
Is it okay if some part of code is shared between two xxxxbsd.rs files? (Maybe something is already shared between macos.rs and bsd.rs)

zonyitoo:
> Is it okay if some part of code is shared (copy-pasted) between two xxxxbsd.rs files? (Maybe something is already shared between macos.rs and bsd.rs)

Your choice. If those features are actually shared between all BSD systems, it could be shared.

ge9:
Now, I 
- separated bsd.rs into freebsd.rs and openbsd.rs
- defined IP_RECVDSTPORT and IPV6_RECVDSTPORT in the global scope of openbsd.rs (as a temporary fix until these two are supported by `libc`)
- added a boolean flag to confirm receiving both the address and port. But this still assumes we receive both the address and port **only once respectively** (in any order). If this is unacceptable, I'll fix it.
- removed `tun2` support

I think I've done all the remaining works (except for the flag issue)

zonyitoo:
Current changes look good to me. But the code are not properly formatted. Run `cargo fmt` in the root directory and push it again.

ge9:
I applied rustfmt, except for the pfvar file. Is this okay?

zonyitoo:
> I applied rustfmt, except for the pfvar file. Is this okay?

It's Ok. But some of them are still not formatted.

ge9:
Oh, sorry, I forgot "git add"...  
I pushed it again. Maybe is it fixed now?

zonyitoo (approved):

PR #1606, opened 2024-08-07
fix(net) losing the raw file descriptor may leak resources

peterwillcn:
```
warning: unused return value of `into_raw_fd` that must be used
  --> crates/shadowsocks/src/net/sys/mod.rs:71:5
   |
71 |     sock.into_raw_fd();
   |     ^^^^^^^^^^^^^^^^^^
   |
   = note: losing the raw file descriptor may leak resources
   = note: `#[warn(unused_must_use)]` on by default
```


zonyitoo:
No bug. Just for suppressing warning.

PR #1616, opened 2024-08-13
feat(local): Allow configuring udp associate addr

vvilhonen:
User might want to bind the socket to a different address than what is advertised in the Socks 5 UDP Associate responses.

This adds a new command line option to control only that but to still default to udp bind addr and local address.

Example:

    sslocal --local-addr 0.0.0.0:1080 --udp-associate-addr 10.51.1.1:1080 -U

Here the server would without this change respond to UDP Associate request with address `0.0.0.0:1080`. With the new switch, it responds `10.51.1.1:1080`.


zonyitoo:
So the UDP_ASSOCIATE command will respond with this `--udp-associate-addr`, but UDP server doesn't listen to this address.

What was the actual usecase in real life?

vvilhonen:
That's correct. For example in kubernetes the services use DNAT to redirect to the actual pods. The pods can't bind to the service's address but they will receive packets addressed to it.

PR #1625, opened 2024-08-26
Add `ProxySocket::as_raw_fd`

dlon:
I need this in order to pass the file descriptor to [`VpnService.protect`](https://developer.android.com/reference/android/net/VpnService#protect(int)).

### DIFF
--- a/crates/shadowsocks/src/relay/udprelay/proxy_socket.rs
+++ b/crates/shadowsocks/src/relay/udprelay/proxy_socket.rs
@@ -27,6 +27,9 @@ use super::crypto_io::{
     ProtocolResult,
 };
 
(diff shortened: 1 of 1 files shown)


dlon:
I've now discovered `vpn_protect_path`! However, in our case, it makes more sense to obtain the fd directly than to pass it over a UDS, and I'm sure there are other use cases as well.

dlon:
@zonyitoo Thank you. Can you please also make a release for this?

zonyitoo:
I am quite busy these days. I will try to make it published ASAP.

zonyitoo:
Released `shadowsocks` v1.20.3.

PR #1634, opened 2024-08-27
fix(readme): update `tun` argument to `tunnel` in `--protocol``

melonges:
```
error: invalid value 'tun' for '--protocol <PROTOCOL>'
[possible values: socks, http, tunnel]
```


zonyitoo:
This is completely wrong. It is `tun` instead of `tunnel`.

Issue #1608, opened 2024-08-11
API - Support Stream+Sink for ProxySocket?


ibigbug:
The tcp stream does take opaque AsyncRead and AsyncWrite as underlying io https://docs.rs/shadowsocks/latest/shadowsocks/relay/tcprelay/proxy_stream/client/struct.ProxyClientStream.html#impl-AsyncRead-for-ProxyClientStream%3CS%3E

thoughts about having the udp socket to take Stream + Sink too ? 

https://docs.rs/shadowsocks/latest/shadowsocks/relay/udprelay/proxy_socket/struct.ProxySocket.html 

so it can be applied to proxy more general udp transports

zonyitoo:
There is no Datagram based sockets in Rust's common libraries supports `Read` / `Write` or `AsyncRead` / `AsyncWrite` , here we are just following the common design in the Rust world.

> so it can be applied to proxy more general udp transports

So what exactly problem are you struggling on?

ibigbug:
there's Sink and Stream https://docs.rs/futures/latest/futures/sink/trait.Sink.html

i'd like to have ss wrap another UDP transport which impls Sink and Stream

zonyitoo:
You can make a wrapper that wraps `ProxyClientStream` and make your own `Sink`

ibigbug:
how to pass it to https://docs.rs/shadowsocks/latest/shadowsocks/relay/udprelay/proxy_socket/struct.ProxySocket.html#method.from_socket

ibigbug:
essentially i want to pass something that is Sink here https://github.com/Watfaq/clash-rs/blob/master/clash_lib/src/proxy/shadowsocks/mod.rs#L234

zonyitoo:
So you want the underlying socket implements `Sink` and let `ProxySocket::from_socket` works with types that implemented `Sink`?

That would require quite a lot of changes in the current implementation.

1. Let `ShadowUdpSocket` implements `Sink`
2. Make all the APIs of `ProxySocket` works with interfaces provided by `Stream` and `Sink`.

I don't see the necessary here because the underlying connection is not a "stream" based connection.

ibigbug:
There's nothing todo with a stream. and I think the Sink and Stream trait is the AsyncRead and AsyncWrite for Udp or any segmented io. 

I know it's a fundamental change that requires amount of work. 

[...]

PR #1641, opened 2024-09-03
take arbitrary IO for udp proxy

ibigbug:
#1608

### DIFF
--- a/crates/shadowsocks/src/relay/udprelay/compat.rs
+++ b/crates/shadowsocks/src/relay/udprelay/compat.rs
@@ -2,6 +2,7 @@ use async_trait::async_trait;
 use std::{
     io::Result,
     net::SocketAddr,
+    ops::Deref,
     task::{Context, Poll},
 };
 use tokio::io::ReadBuf;
@@ -24,6 +25,11 @@ pub trait DatagramTransport: Send + Sync + std::fmt::Debug {
     fn poll_send(&self, cx: &mut Context<'_>, buf: &[u8]) -> Poll<Result<usize>>;
     fn poll_send_to(&self, cx: &mut Context<'_>, buf: &[u8], target: SocketAddr) -> Poll<Result<usize>>;
     fn poll_send_ready(&self, cx: &mut Context<'_>) -> Poll<Result<()>>;
+
+    fn local_addr(&self) -> Result<SocketAddr>;
+
+    #[cfg(unix)]
+    fn as_raw_fd(&self) -> std::os::fd::RawFd;
 }
 
 #[async_trait]
@@ -53,7 +59,7 @@ impl DatagramTransport for UdpSocket {
     }
 
     fn poll_recv_ready(&self, cx: &mut Context<'_>) -> Poll<Result<()>> {
-        UdpSocket::poll_recv_ready(self, cx)
+        self.deref().poll_recv_ready(cx)
     }
 
     fn poll_send(&self, cx: &mut Context<'_>, buf: &[u8]) -> Poll<Result<usize>> {
@@ -65,6 +71,18 @@ impl DatagramTransport for UdpSocket {
     }
 
     fn poll_send_ready(&self, cx: &mut Context<'_>) -> Poll<Result<()>> {
-        UdpSocket::poll_send_ready(self, cx)
+        self.deref().poll_send_ready(cx)
+    }
+
+    fn local_addr(&self) -> Result<SocketAddr> {
+        self.deref().local_addr()
+    }
+
+    #[cfg(unix)]
+    fn as_raw_fd(&self) -> std::os::fd::RawFd {
+        use std::ops::Deref;
+        use std::os::fd::AsRawFd;
+
+        self.deref().as_raw_fd()
     }
 }
--- a/crates/shadowsocks/src/relay/udprelay/compat.rs
+++ b/crates/shadowsocks/src/relay/udprelay/compat.rs
@@ -1,0 +1,70 @@
+use async_trait::async_trait;
+use std::{
+    io::Result,
+    net::SocketAddr,
+    task::{Context, Poll},
+};
+use tokio::io::ReadBuf;
+
+use crate::net::UdpSocket;
+
+/// a trait for datagram transport that wraps around a tokio `UdpSocket`
+#[async_trait]
+pub trait DatagramTransport: Send + Sync + std::fmt::Debug {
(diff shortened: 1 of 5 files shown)


zonyitoo:
Why was that necessary? `SocketAddr` is also implements `ToSocketAddrs`.

ibigbug:
We need it to be SocketAddr when passing it to the underlying udp send

zonyitoo:
> Why was that necessary? `SocketAddr` is also implements `ToSocketAddrs`.

What was the key issue of the current master implementation that blocked you from passing `SocketAddr` to the APIs?

ibigbug:
There is no issue passing a SocketAddr, but we can't and shouldn't pass IntoSocketAddrs into an abstract udp up, which is not always a tokio udpsocket

ibigbug:
just realised it's a bit tricky to have the abstract io to be compatible with the raw socket operations. maybe I should separate two kinds of io and fail all the raw socket operation on the abstract one and give the control of that back to the author of which

zonyitoo:
Please resolve all the CI failures.

ibigbug:
Sure. There are still more to do. Will update soon.

zonyitoo:
I was wondering an other solution:

1. Make a `trait DatagramTransport`, which has APIs: `send`, `recv`, `send_to`, `recv_from`.
2. `ProxySocket` took a type impls `DatagramTransport` as the underlying I/O socket
3. `UdpSocket` impls `DatagramTransport`, which will require no `lock` or any other synchronization method in the application level.
4. You can impls your customized transport by impls `DatagramTransport` to support `Sink` and `Stream`.

crates/shadowsocks/src/relay/udprelay/proxy_socket.rs
```diff
@@ -69,10 +72,12 @@ impl From<ProxySocketError> for io::Error {
 pub type ProxySocketResult<T> = Result<T, ProxySocketError>;
 
 /// UDP client for communicating with ShadowSocks' server
+/// Thread safety: users of `ProxySocket` should not share it across threads
+// TODO: unless a proper spinlock on `.socket` is implemented
```
ibigbug, line 76:
```suggestion
```

crates/shadowsocks/src/relay/udprelay/compat.rs
```diff
@@ -0,0 +1,88 @@
...
+use crate::net::UdpSocket;
+
+/// a trait for datagram transport that wraps around a tokio `UdpSocket`
+#[async_trait]
+pub trait DatagramTransport: Send + Sync + std::fmt::Debug {
+    async fn recv(&self, buf: &mut [u8]) -> Result<usize>;
+    async fn recv_from(&self, buf: &mut [u8]) -> Result<(usize, SocketAddr)>;
+
+    async fn send(&self, buf: &[u8]) -> Result<usize>;
+    async fn send_to(&self, buf: &[u8], target: SocketAddr) -> Result<usize>;
```
zonyitoo, line 19: Since we already have `poll_*` functions, these `async` wrappers could be added in an `DatagramTransportExt` trait.

crates/shadowsocks/src/relay/udprelay/proxy_socket.rs
```diff
@@ -72,7 +75,7 @@ pub type ProxySocketResult<T> = Result<T, ProxySocketError>;
 #[derive(Debug)]
 pub struct ProxySocket {
     socket_type: UdpSocketType,
-    socket: ShadowUdpSocket,
+    io: Box<dyn DatagramTransport>,
```
zonyitoo, line 78: Why not just make `io` to be a generic parameter

```rust
pub struct ProxySocket<S>
    where S: DatagramTransport
{
    socket: S
}
```

crates/shadowsocks/src/relay/udprelay/compat.rs
```diff
@@ -0,0 +1,88 @@
...
+    ops::Deref,
+    task::{Context, Poll},
+};
+use tokio::io::ReadBuf;
+
+use crate::net::UdpSocket;
+
+/// a trait for datagram transport that wraps around a tokio `UdpSocket`
+#[async_trait]
+pub trait DatagramTransport: Send + Sync + std::fmt::Debug {
```
zonyitoo, line 14: Are these bounded traits has to be `Send`, `Sync` and `Debug`?

It could be restricted on `ProxySocket<S>`.

crates/shadowsocks/src/relay/udprelay/proxy_socket.rs
zonyitoo: As mentioned above, `S` could be `S: DatagramTransport + Send + Sync + Debug`

zonyitoo (changes requested):

crates/shadowsocks/src/relay/udprelay/proxy_socket.rs
ibigbug: i think that'd make the ProxySocket to be a generic, which can propgate to the entire call chain.

but I prefer it too, which would be static dispatch instead of dynamic

ibigbug:
yeah of course, feel free to update

crates/shadowsocks/src/relay/udprelay/compat.rs
ibigbug: yes but to use it, one has to impl Sync Send anyways

crates/shadowsocks/src/relay/udprelay/compat.rs
ibigbug: yes true, but thing is it won't work with ProxySocket without implementing all of these methods.

so i doubt spliting the Ext trait

zonyitoo:
@ibigbug Please set your branch to be editable by maintainer.

ibigbug:
@zonyitoo i odn't see that option, maybe it's from a forked repo?

zonyitoo:
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

> On user-owned forks, if you want to allow anyone with push access to the upstream repository to make changes to your pull request, select Allow edits from maintainers.

ibigbug:
image
 nothing shows up here?

ibigbug:
https://github.com/orgs/community/discussions/5634

seems it doesn't work for org

Issue #1697, opened 2024-10-02
Debian package: Multiple warnings


marek22k:
Hello,

there seem to be several warnings when building the Debian package:
```
dpkg-buildpackage: warning:     debian/changelog(l8): found start of entry where expected more change data or trailer
dpkg-buildpackage: warning:     debian/changelog(l8): found end of file where expected more change data or trailer
dpkg-source: warning: shadowsocks-rust/debian/changelog(l8): found start of entry where expected more change data or trailer
dpkg-source: warning: shadowsocks-rust/debian/changelog(l8): found end of file where expected more change data or trailer
dh: warning: Compatibility levels before 10 are deprecated (level 9 in use)
dh_auto_clean: warning: Compatibility levels before 10 are deprecated (level 9 in use)
dh_clean: warning: Compatibility levels before 10 are deprecated (level 9 in use)
dh: warning: Compatibility levels before 10 are deprecated (level 9 in use)
dh_auto_configure: warning: Compatibility levels before 10 are deprecated (level 9 in use)
dh_auto_build: warning: Compatibility levels before 10 are deprecated (level 9 in use)
dh_auto_test: warning: Compatibility levels before 10 are deprecated (level 9 in use)
dh_auto_install: warning: Compatibility levels before 10 are deprecated (level 9 in use)
dh_install: warning: Compatibility levels before 10 are deprecated (level 9 in use)
dh_installdocs: warning: Compatibility levels before 10 are deprecated (level 9 in use)
dh_installchangelogs: warning:     debian/changelog(l8): found start of entry where expected more change data or trailer
dh_installchangelogs: warning:     debian/changelog(l8): found end of file where expected more change data or trailer
dh_installchangelogs: warning: Could not parse timestamp '16 Nov 00:11:44 2021 +0800'. debian/changelog will not be trimmed.
dh_installchangelogs: warning: debian/changelog could not be trimmed. The full changelog will be installed.
dh_installinit: warning: Compatibility levels before 10 are deprecated (level 9 in use)
dh_compress: warning: Compatibility levels before 10 are deprecated (level 9 in use)
dh_missing: warning: Compatibility levels before 10 are deprecated (level 9 in use)
dh_strip: warning: Compatibility levels before 10 are deprecated (level 9 in use)
dh_makeshlibs: warning: Compatibility levels before 10 are deprecated (level 9 in use)
dh_shlibdeps: warning: Compatibility levels before 10 are deprecated (level 9 in use)
[...]

PR #1722, opened 2024-10-20
debian: fixup debian packaging build scripts

mesher2024:
1. Fix up multiple warnings caused by compact level (#1697).
2. Replace nobody user in shadowsocks-rust.service unit file with DynamicUser, add DynamicUser in shadowsocks-rust-server@.service and shadowsocks-rust-local@.service unit file.
3. Fix up postinst Dash script error caused by Bash string manipulation, replace it with sed script.


mesher2024:
4. Remove dependency on pwgen in debian/control (#1702 ).

PR #1790, opened 2024-11-29
Allow IP fragmentation for outbound UDP sockets

### LINKED ISSUE #1780: Configuration option to allow IP fragmentation on outbound UDP sockets in ssserver
Hello! I have a use case where outbound UDP traffic from ssserver sometimes needs to traverse a network with a smaller MTU. Unfortunately I don't have control over the network or the application sending the traffic.

I saw that per previous issues on this project, IP fragmentation was disabled to improve performance. For my use case, it's preferable to allow IP fragmentation / accept the performance overhead.

I tested that by commenting out the following lines, UDP traffic was fragmented as expected: 
https://github.com/shadowsocks/shadowsocks-rust/blob/e791f233f7848d6c9f2a99b0794986d0cddb6c79/crates/shadowsocks/src/net/sys/unix/linux/mod.rs#L313-L315

What do you think about adding a new configuration option in ssserver like `allow_outbound_udp_fragmentation` ? The default will be false, keeping the current behavior. When `allow_outbound_udp_fragmentation = true`, `set_disable_ip_fragmentation` will be skipped. If that sounds OK, I'll be happy to submit a patch.

tom-ryan:
Fixes #1780

### DIFF
--- a/crates/shadowsocks/src/net/option.rs
+++ b/crates/shadowsocks/src/net/option.rs
@@ -65,6 +65,9 @@ pub struct ConnectOpts {
     /// Outbound socket binds to interface
     pub bind_interface: Option<String>,
 
+    /// Outbound UDP socket allows IP fragmentation
+    pub udp_allow_fragmentation: bool,
+
     /// TCP options
     pub tcp: TcpSocketOpts,
 
--- a/crates/shadowsocks/src/net/sys/unix/bsd/freebsd.rs
+++ b/crates/shadowsocks/src/net/sys/unix/bsd/freebsd.rs
@@ -251,8 +251,10 @@ pub async fn bind_outbound_udp_socket(bind_addr: &SocketAddr, _config: &ConnectO
         UdpSocket::from_std(socket.into())?
     };
 
-    if let Err(err) = set_disable_ip_fragmentation(af, &socket) {
-        warn!("failed to disable IP fragmentation, error: {}", err);
+    if ! config.udp_allow_fragmentation {
+        if let Err(err) = set_disable_ip_fragmentation(af, &socket) {
+            warn!("failed to disable IP fragmentation, error: {}", err);
+        }
     }
 
     Ok(socket)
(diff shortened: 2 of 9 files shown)


crates/shadowsocks/src/net/option.rs
```diff
@@ -65,6 +65,9 @@ pub struct ConnectOpts {
     /// Outbound socket binds to interface
     pub bind_interface: Option<String>,
 
+    /// Outbound UDP socket allows IP fragmentation
+    pub udp_allow_fragmentation: bool,
```
tom-ryan, line 69: Added the field on `ConnectOpts` [as suggested](https://github.com/shadowsocks/shadowsocks-rust/issues/1780#issuecomment-2497174945) since `UdpSocketOpts` is used for both inbound and outbound sockets, but we could move it to `UdpSocketOpts` instead.

zonyitoo: Ah, it should be more reasonable to be in `UdpSocketOpts`.

crates/shadowsocks/src/net/sys/unix/bsd/freebsd.rs
```diff
@@ -251,8 +251,10 @@ pub async fn bind_outbound_udp_socket(bind_addr: &SocketAddr, _config: &ConnectO
         UdpSocket::from_std(socket.into())?
     };
 
-    if let Err(err) = set_disable_ip_fragmentation(af, &socket) {
-        warn!("failed to disable IP fragmentation, error: {}", err);
+    if ! config.udp_allow_fragmentation {
```
zonyitoo, line 254: Emm, the space after `!` is not needed. You should run `cargo fmt` before commit.

tom-ryan:
Thanks @zonyitoo . Do you want me to add a commit to address your comments?

Do you know yet when the next release will be?

zonyitoo:
> Thanks @zonyitoo . Do you want me to add a commit to address your comments?
> 
> Do you know yet when the next release will be?

I have already add them in the master branch.

Next release is blocking on hickory-dns's next release. They said it would be early December.

tom-ryan:
> > Thanks @zonyitoo . Do you want me to add a commit to address your comments?
> > Do you know yet when the next release will be?
> 
> I have already add them in the master branch.
> 
> Next release is blocking on hickory-dns's next release. They said it would be in early December.

Great! Thank you

PR #1820, opened 2024-12-21
chore: allow to export pub socks handlers

estin:
Hi! It's minor changes to allow wrap tcp stream with socks logic (relay)


zonyitoo:
Why?

zonyitoo:
The server sends SOCKS 5 handshakes like a client?

zonyitoo:
Well, this doesn't seem to be a common use case of this Project. I don't want to make these modules `pub` because they are all internal implementation details, I don't want to maintain compatibility in future updates.

PR #1834, opened 2025-01-06
refactor: change error handling to be strongly-typed

ClaytonKnittel:
Hi! New contributor to this project here.

This PR refactors error handling in the local/manager/server binaries, resulting in a net reduction of 26 lines of code.

### DIFF
--- a/bin/sswinservice.rs
+++ b/bin/sswinservice.rs
@@ -1,14 +1,16 @@
 use std::{
     ffi::OsString,
     future::Future,
-    process::ExitCode,
     sync::atomic::{AtomicU32, Ordering},
     time::Duration,
 };
 
 use clap::Command;
 use log::{error, info};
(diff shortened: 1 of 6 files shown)


ClaytonKnittel:
I'm unsure if this is a breaking change, this is modifying the signatures of `pub` functions in `src/` and removing the `pub const` error codes. Feedback is welcome!

zonyitoo:
LGTM. Please add doc comments for structs, methods and the mod itself.

zonyitoo:
And what's the actual use-case of the error type? Do you actually use it as a library?

ClaytonKnittel:
Done.

The error type is not intended to be used outside of the code I've modified. I've changed the package visibility of mod error to package-local.

zonyitoo:
@ClaytonKnittel Please fix windows build.

ClaytonKnittel:
Fixed.

PR #1922, opened 2025-03-20
tokio_util::sync::CancellationToken usage in local-tun

ssrlive:


crates/shadowsocks-service/src/local/tun/tcp.rs
```diff
@@ -469,24 +460,22 @@ impl TcpTun {
                                 .poll_delay(before_poll, &socket_set)
                                 .unwrap_or(SmolDuration::from_millis(5));
                             if next_duration != SmolDuration::ZERO {
-                                thread::park_timeout(Duration::from(next_duration));
+                                std::thread::park_timeout(Duration::from(next_duration));
```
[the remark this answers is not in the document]
ssrlive: So, how to resolve it?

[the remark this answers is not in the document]
ssrlive: I have change `socket_creation_rx.try_recv()` to `socket_creation_rx.recv()`.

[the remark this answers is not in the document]
ssrlive: It means that should remove some `manager_notify.cancel()` calls.

crates/shadowsocks-service/src/local/tun/tcp.rs
```diff
@@ -469,24 +454,22 @@ impl TcpTun {
                                 .poll_delay(before_poll, &socket_set)
                                 .unwrap_or(SmolDuration::from_millis(5));
                             if next_duration != SmolDuration::ZERO {
-                                thread::park_timeout(Duration::from(next_duration));
+                                tokio::time::sleep(Duration::from(next_duration)).await;
```
zonyitoo, line 457: It may not be a good choice to make this loop running as a Task running in tokio's runtime, because it will use most of the CPU time of worker threads in the runtime, which will eventually starve the other tasks.

And this should not be changed to `sleep`. I use `park_timeout` here because we can use `unpark()` to wake up the thread and continue the loop. If it changed to `sleep`, then `poll_delay` will not be called immediately after sockets' status changed.

crates/shadowsocks-service/src/local/tun/tcp.rs
```diff
@@ -203,10 +185,7 @@ impl AsyncWrite for TcpConnection {
 
         let n = control.send_buffer.enqueue_slice(buf);
 
-        if n > 0 {
-            self.manager_notify.notify();
-        }
-        Ok(n).into()
+        if n > 0 { Ok(n).into() } else { Poll::Pending }
```
zonyitoo, line 188: After data was copied into the `send_buffer`, we should wake up the `poll` thread to copy data from `send_buffer` into socket's send buffer, otherwise delay would be increased.

crates/shadowsocks-service/src/local/tun/tcp.rs
ssrlive: After reviewing it, just modify it directly on my fork.

zonyitoo: Your commits are in this PR, if I haven't missed any other changes. This comment is based on your latest commit.

zonyitoo:
Have you run a stress test with your changes? Current implementation of local-tun couldn't reach 1000Mbps in one connection, the most significant bottleneck is the loop thread, which will become too busy when there are too many data transferred.

I don't think the thread notifier is the key to resolve this bottleneck.

ssrlive:
I face the same problem. When I use ipstack components for tunnel
transmission, the big data volume may be stuck. So I will investigate the
code implementation of your tunnel.

zonyitoo:
I have been looking around for the best solution about high performance implemenation with smoltcp and tokio.

1. `tokio-smoltcp`, a very straight forward implementation, lock the whole world when any of the sockets want I/O.
2. `onetun`, a WireGuard implemenation. It copies all data into a global queue (called bus).
3. ...

The key problem to make smoltcp to work properly with tokio is that smoltcp requires all sockets to be managed by `SocketSet`, and `Interface`'s API `poll` will check every sockets' status in the `SocketSet` on each loop. The `SocketSet` has to be `mut`, so it can only to be shared with a `Mutex` and `lock()` the whole world when any tokio Task wants to mutate any of the sockets in the `SocketSet`, or use separate send/recv buffers in tokio's Tasks, which are guarded by many  lightweight Spin locks. The latter solution is the one I chose in this Project, which significantly reduce lock contention. This makes single connection brand-width reaches 600Mbps, but far from 1000Mbps.

zonyitoo:
I have this idea before, but haven’t actually implemented it yet. As I said, when I run stress test with one single connection, brand-width is still low. In this case, there was only 1 socket in the SocketSet.

ssrlive:
https://github.com/smoltcp-rs/smoltcp/blob/main/src/socket/tcp/congestion.rs#L37-L45

smoltcp has congestion control implemented, did you enable it?

zonyitoo:
No. Thanks for your advice. Let me try `cubic` right now.

This PR is going to be closed. I think this won't help with improving performance.

ssrlive:
What are the results of stress testing local-tun after enabling congestion control?

zonyitoo:
Haven't tested it yet. I may do it in this weekend. If you have time, please do me a flavor.

PR #1929, opened 2025-04-03
Apply outbound_bind_addr to UDP traffic in ssserver

### LINKED ISSUE #1928: IPv4 outbound_bind_addr not being applied for UDP traffic in ssserver
We are using shadowsocks to proxy both TCP and UDP traffic. We want to use the outbound_bind_addr to send outbound traffic from ssserver from a different IP address. We are only using IPv4 in our environment.

We are seeing an issue where outbound UDP traffic from ssserver does not use the outbound_bind_addr when outbound_bind_addr is an IPv4 address. Instead, outbound UDP traffic uses the ssserver server IP. Outbound TCP traffic does use the outbound_bind_addr correctly.

It looks like this is happening because [when a UDP socket is created](https://github.com/shadowsocks/shadowsocks-rust/blob/master/crates/shadowsocks/src/net/sys/unix/linux/mod.rs#L287), it expects the AddrFamily and the SocketAddr to be the same type for the bind_addr to be applied. But for UDP it looks like the AddrFamily is always IPv6, even though the outbound_bind_addr is IPv4. This causes the match statement for bind_addr assignment to not apply the configured outbound_bind_addr and it defaults to the ssserver server IP instead. 

In our testing, if we use an IPv4-mapped IPv6 address for the outbound_bind_addr, UDP traffic successfully originates from the IPv6 outbound_bind_addr but TCP traffic is unsuccessful.

How can we enable an IPv4 outbound_bind_addr for TCP and UDP traffic?

An idea could be to make dual stack configuration configurable as a ssserver config option, so using only IPv4 or only IPv6 are options. If you can suggest a solution, we’re happy to push up a patch.

luujenn:
Fixes #1928

### DIFF
--- a/crates/shadowsocks/src/net/sys/unix/bsd/freebsd.rs
+++ b/crates/shadowsocks/src/net/sys/unix/bsd/freebsd.rs
@@ -231,10 +231,7 @@ pub async fn create_outbound_udp_socket(af: AddrFamily, config: &ConnectOpts) ->
         (AddrFamily::Ipv4, Some(SocketAddr::V6(addr))) => {
             // Map IPv6 bind_local_addr to IPv4 if AF is IPv4
             match addr.ip().to_ipv4_mapped() {
-                Some(addr) => {
-                    let ip_addr: IpAddr = addr.into();
-                    SocketAddr::new(ip_addr, 0)
-                },
+                Some(addr) => SocketAddr::new(IpAddr::from(addr), 0),
                 None => return Err(Error::new(ErrorKind::InvalidInput, "Invalid IPv6 address")),
             }
         },
--- a/crates/shadowsocks/src/net/sys/unix/bsd/freebsd.rs
+++ b/crates/shadowsocks/src/net/sys/unix/bsd/freebsd.rs
@@ -1,6 +1,6 @@
 use std::{
-    io, mem,
-    net::{Ipv4Addr, Ipv6Addr, SocketAddr},
+    io, mem, io:ErrorKind,
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
     os::unix::io::{AsRawFd, RawFd},
     pin::Pin,
     ptr,
@@ -19,7 +19,7 @@ use tokio_tfo::TfoStream;
 
 use crate::net::{
     AcceptOpts, AddrFamily, ConnectOpts,
-    sys::{set_common_sockopt_after_connect, set_common_sockopt_for_connect, socket_bind_dual_stack},
+    sys::{set_common_sockopt_after_connect, set_common_sockopt_for_connect, socket_bind_dual_stack, io::Error},
     udp::{BatchRecvMessage, BatchSendMessage},
 };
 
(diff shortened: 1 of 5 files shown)


crates/shadowsocks/src/net/sys/unix/bsd/freebsd.rs
```diff
@@ -228,7 +228,22 @@ pub fn set_disable_ip_fragmentation<S: AsRawFd>(af: AddrFamily, socket: &S) -> i
 pub async fn create_outbound_udp_socket(af: AddrFamily, config: &ConnectOpts) -> io::Result<UdpSocket> {
     let bind_addr = match (af, config.bind_local_addr) {
         (AddrFamily::Ipv4, Some(SocketAddr::V4(addr))) => addr.into(),
+        (AddrFamily::Ipv4, Some(SocketAddr::V6(addr))) => {
+            // Map IPv6 bind_local_addr to IPv4 if AF is IPv4
+            match addr.ip().to_ipv4_mapped() {
+                Some(addr) => {
+                    let ip_addr: IpAddr = addr.into();
```
zonyitoo, line 235: So type inferation could work in here? You could change it to `SocketAddr::new(IpAddr::from(addr), 0)`

luujenn: I'll make the change

luujenn: Sorry I misread your comment, I think it complained about the functions not being available for different socket types so the build was unsuccessful. But I'll double check that

luujenn: @zonyitoo Inferation worked as expected, I pushed up the change

crates/shadowsocks/src/net/sys/unix/bsd/freebsd.rs
```diff
@@ -228,7 +228,19 @@ pub fn set_disable_ip_fragmentation<S: AsRawFd>(af: AddrFamily, socket: &S) -> i
...
+            // Map IPv6 bind_local_addr to IPv4 if AF is IPv4
+            match addr.ip().to_ipv4_mapped() {
+                Some(addr) => SocketAddr::new(IpAddr::from(addr), 0),
+                None => return Err(Error::new(ErrorKind::InvalidInput, "Invalid IPv6 address")),
+            }
+        },
         (AddrFamily::Ipv6, Some(SocketAddr::V6(addr))) => addr.into(),
+        (AddrFamily::Ipv6, Some(SocketAddr::V4(addr))) => {
+            // Map IPv4 bind_local_addr to IPv6 if AF is IPv6
+            let ip_addr: IpAddr = addr.ip().to_ipv6_mapped().into();
```
zonyitoo, line 241: It could also be applied to this change, `SocketAddr::new(IpAddr::from(addr.ip().to_ipv6_mapped()), 0)`

luujenn: Sorry I missed this. I can push up a commit with this suggestion, if you'd like?
Do you know yet when the next release will be?

zonyitoo: I have already fixed it in master. Some update builds are failed when upgrading dependencies, I may make a new release after resolving them.

PR #1968, opened 2025-06-12
fix: make hickory-dns truly optional

AaronChen0:
`hickory-dns` is enabled by default even using `--no-default-features`, though it is specified as optional in README.


AaronChen0:
Hi, just a friendly ping to check in on this PR. Let me know if there's anything I can do. Thanks!

PR #1972, opened 2025-06-15
Allow user to pass SocketProtector callback function for android

zhangsan946:
Allow user to implement the Rust interface which calls the Java VPNService protect method, and to be used to protect the socket fd right after the socket is created on Android.

### DIFF
--- a/crates/shadowsocks/src/net/mod.rs
+++ b/crates/shadowsocks/src/net/mod.rs
@@ -11,7 +11,7 @@ pub use self::{
     udp::UdpSocket,
 };
 #[cfg(target_os = "android")]
-pub use self::option::android::SocketProtector;
+pub use self::option::android::{CloneFn, SocketProtectFn, socket_protect_fn};
 
 mod option;
 mod sys;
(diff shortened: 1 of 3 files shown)


zonyitoo:
Should `SocketProtector` defined as a `trait`?

zhangsan946:
@zonyitoo there is no need, when user wanna pass a callback, all he needs to do is to create a `SocketProtector` with `SocketProtector::new(move |fd| callback(fd))` and pass to `ConnectOpts`. If to be trait, it is expected user needs implement a dedicated struct just for this callback while the former is a much simpler implementation.

zonyitoo:
No. Users don't need to really implement this trait if they only has a function, just add a helper function:

```rust
fn make_socket_protector<F>(f: F) -> FnSocketProtector<F> ...
```

which `FnSocketProtector<F>` is exactly the same as your `SocketProtector` implementation.

This design could be found in `tower`, `hyper`, ... which could allow users who doesn't want to simply call a function, they could make a "Functor" with states.

PR #1974, opened 2025-06-17
Allow user to pass SocketProtectFn for android

zhangsan946:
Allow user to implement the Rust interface which calls the Java VPNService protect method on Android, and to be used to protect the socket fd right after the socket is created.

### DIFF
--- a/crates/shadowsocks/src/net/mod.rs
+++ b/crates/shadowsocks/src/net/mod.rs
@@ -4,6 +4,8 @@ use std::net::SocketAddr;
 
 #[cfg(unix)]
 pub use self::sys::uds::{UnixListener, UnixStream};
+#[cfg(target_os = "android")]
+pub use self::option::android::{SocketProtect, SocketProtectFn};
 pub use self::{
     option::{AcceptOpts, ConnectOpts, TcpSocketOpts, UdpSocketOpts},
     sys::{IpStackCapabilities, get_ip_stack_capabilities, set_tcp_fastopen, socket_bind_dual_stack},


zhangsan946:
@zonyitoo, Please check this new PR.

crates/shadowsocks/src/net/option.rs
```diff
@@ -87,3 +89,56 @@ pub struct AcceptOpts {
...
+    where 
+        F: Fn(i32)
+    {
+        pub fn call(&self, fd: i32) {
+            (self.f)(fd)
+        }
+    }
+
+    impl Clone for SocketProtectFn<Box<dyn CloneFn>>
+    {
```
zonyitoo, line 131: Please run `cargo fmt`.

crates/shadowsocks/src/net/option.rs
```diff
@@ -87,3 +89,56 @@ pub struct AcceptOpts {
     /// Enable IPV6_V6ONLY option for socket
     pub ipv6_only: bool,
 }
+
+#[cfg(target_os = "android")]
+pub mod android {
+    pub fn socket_protect_fn<F>(f: F) -> SocketProtectFn<Box<dyn CloneFn>>
```
zonyitoo, line 95: Public APIs should have doucuments.

crates/shadowsocks/src/net/option.rs
```diff
@@ -87,3 +89,56 @@ pub struct AcceptOpts {
...
+        fn clone_box(&self) -> Box<dyn CloneFn> {
+            Box::new(self.clone())
+        }
+    }
+
+    pub struct SocketProtectFn<F> {
+        f: F,
+    }
+
+    impl<F> SocketProtectFn<F>
```
zonyitoo, line 121: Consider renamed to `SocketProtect`

zonyitoo:
This is not the design I have mentioned in the previous PR. What I wanted was:

```rust
trait SocketProtect {
    fn protect(fd: RawFd) -> io::Result<()>;
}

struct SocketProtectFn<F> {
    f: F
}

impl<F> SocketProtect for SocketProtectFn<F>
where F: Fn(RawFd) -> io::Result<()>
{
    fn protect(&self, fd: RawFd) -> io::Result<()> {
        self.f(fd)
    }
}

pub fn make_socket_protect_fn(f: F) -> SocketProtectFn<F>
where F: Fn(RawFd) -> io::Result<()> + Send + Sync + Clone
{
    SocketProtectFn { f }
}

struct ConnectOpts {
    // ... other fields

    #[cfg(target_os = "android")]
    pub vpn_socket_protect_fn: Option<Arc<Box<dyn SocketProtect + Send + Sync>>>,
}
```

I would recommend `Arc<Box<dyn SocketProtect>>`. The function itself doesn't need to be clonable.

crates/shadowsocks/src/net/option.rs
```diff
@@ -97,33 +97,33 @@ pub mod android {
     use std::os::unix::io::RawFd;
     use std::sync::Arc;
 
-    pub fn make_socket_protect_fn<F>(f: F) -> SocketProtectFn<Arc<Box<F>>>
+    pub fn make_socket_protect_fn<F>(f: F) -> SocketProtectFn<Arc<Box<dyn SocketProtect + Send + Sync>>>
```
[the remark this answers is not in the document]
zhangsan946: I am struggling to make it work right now.. and will let you know once it is ready for review again..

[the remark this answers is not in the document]
zonyitoo: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=8c395f085df918c902d91861b9f40e05

This is a working example about my original idea.

PR #1975, opened 2025-06-18
Add and expose make_socket_protect_fn

zhangsan946:
Hi @zonyitoo, thanks for your patience and guidence for the last PR. I made some minor change to make it more flexbile for the end user to config `vpn_socket_protect`.


zonyitoo:
You don't need to do that actually. I added `MakeSocketProtect` trait in the last commit, it would allow users to set by `set_vpn_socket_protect` with any types that implements `MakeSocketProtect`. So a standalone `make_socket_protect_fn` is not needed.

PR #1983, opened 2025-07-04
fix(shadowsocks-service): fix compilation without `aead-cipher`

### LINKED ISSUE #1982: 1.23.5 does not compile with `aead-cipher` feature disabled
I'm trying to build with such flags:
```
$ cargo build --release --bin ssservice --no-default-features --features 'service,logging,multi-threaded,aead-cipher-2022'
   Compiling shadowsocks-service v1.23.5 (/tmp/shadowsocks-rust/crates/shadowsocks-service)
error[E0308]: mismatched types
   --> crates/shadowsocks-service/src/manager/server.rs:482:43
    |
482 |             None => return Ok(AddResponse("method is required")),
    |                               ----------- ^^^^^^^^^^^^^^^^^^^^- help: try using a conversion method: `.to_string()`
    |                               |           |
    |                               |           expected `String`, found `&str`
    |                               arguments to this struct are incorrect
    |
note: tuple struct defined here
   --> /tmp/shadowsocks-rust/crates/shadowsocks/src/manager/protocol.rs:74:12
    |
74  | pub struct AddResponse(pub String);
    |            ^^^^^^^^^^^

For more information about this error, try `rustc --explain E0308`.
error: could not compile `shadowsocks-service` (lib) due to 1 previous error
```
Rust compiler is up to date:
```
$ rustc --version
rustc 1.88.0 (6b00bc388 2025-06-23) (Arch Linux rust 1:1.88.0-1)
```
Previous versions compiled just fine.

HanabishiRecca:
#1968 made it possible to actually disable `aead-cipher` feature for `shadowsocks-service`.
That exposed previously unreachable code path which does not compile.

Closes #1982.

### DIFF
--- a/crates/shadowsocks-service/src/manager/server.rs
+++ b/crates/shadowsocks-service/src/manager/server.rs
@@ -479,7 +479,7 @@ impl Manager {
             #[cfg(feature = "aead-cipher")]
             None => self.svr_cfg.method.unwrap_or(CipherKind::CHACHA20_POLY1305),
             #[cfg(not(feature = "aead-cipher"))]
-            None => return Ok(AddResponse("method is required")),
+            None => return Ok(AddResponse("method is required".to_string())),
         };
 
         let mut svr_cfg = match ServerConfig::new(addr, req.password.clone(), method) {


zonyitoo:
Thanks for contributing. Does it require making a new release? I wouldn't make a new release with only this fix if it didn't affect most users.

zonyitoo:
Anyway, making a new release for shadowsocks-service.

PR #1992, opened 2025-07-25
feat: support file logging with tracing

shadowsocks69420:
### Summary

Enable file logging with tracing via `tracing-appender` and introduce new `log.file.*` configuration options.

As `log4rs` is scheduled for removal, this PR serves as an intermediate step toward that transition. I’ve also removed user-facing documentation pertaining to `log4rs` to reduce the impact of potential future breaking changes.

#### Console output will be suppressed if file logging is configured.

Significant additional work is required to support both.

#### Does not support rotation based on file size.

This is an inherent limitation of the `tracing-appender` library.

#### Plugin output will not be written to file

This is an existing issue that applies to `log4rs` as well.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -142,29 +142,29 @@ impl Config {
                 nlog.format = nformat;
             }
 
-            if let Some(file_config) = log.file
+            if let Some(file_config) = log.file {
                 // directory must be configured for file logging
-                && let Some(directory) = file_config.directory
-            {
-                let mut nfile = LogFileConfig::new(directory);
-                if let Some(rotation) = file_config.rotation {
-                    nfile.rotation = match rotation.as_str() {
-                        "never" => tracing_appender::rolling::Rotation::NEVER,
-                        "hourly" => tracing_appender::rolling::Rotation::HOURLY,
-                        "daily" => tracing_appender::rolling::Rotation::DAILY,
-                        _ => return Err(ConfigError::InvalidValue(rotation)),
-                    };
+                if let Some(directory) = file_config.directory {
+                    let mut nfile = LogFileConfig::new(directory);
+                    if let Some(rotation) = file_config.rotation {
+                        nfile.rotation = match rotation.as_str() {
+                            "never" => tracing_appender::rolling::Rotation::NEVER,
+                            "hourly" => tracing_appender::rolling::Rotation::HOURLY,
+                            "daily" => tracing_appender::rolling::Rotation::DAILY,
+                            _ => return Err(ConfigError::InvalidValue(rotation)),
+                        };
+                    }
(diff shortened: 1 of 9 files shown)


src/config.rs
```diff
@@ -142,6 +142,31 @@ impl Config {
...
+                }
+                if let Some(prefix) = file_config.prefix {
+                    nfile.prefix = Some(prefix);
+                }
+                if let Some(suffix) = file_config.suffix {
+                    nfile.suffix = Some(suffix);
+                }
+                if let Some(max_files) = file_config.max_files {
+                    nfile.max_files = Some(max_files);
+                }
```
shadowsocks69420, line 166: There is a small caveat FWIW: if the user sets `max_files` to `0`, it will cause panicking in debug builds due to integer underflow in the `tracing-appender` crate[^1].
However, this doesn't cause any noticeable issue if using the release build as the check will be omitted and it behaves just like `u64::MAX`. Furthermore, I strongly believe the `tracing-appender` crate shouldn't accept `0` here in the first place. Therefore, no zero check is performed here.

[^1]: https://github.com/tokio-rs/tracing/blob/e63ef57f3d686abe3727ddd586eb9af73d6715b7/tracing-appender/src/rolling.rs#L695

shadowsocks69420 (commented):
Pointing out some changes that may worth discussing.

src/config.rs
```diff
@@ -142,6 +142,31 @@ impl Config {
                 nlog.format = nformat;
             }
 
+            if let Some(file_config) = log.file
+                // directory must be configured for file logging
+                && let Some(directory) = file_config.directory
+            {
```
shadowsocks69420, line 148: This requires [rustc 1.88.0](https://blog.rust-lang.org/2025/06/26/Rust-1.88.0/#let-chains). Let me know if I should bump the MSRV or nest this block one level deeper.

shadowsocks69420:
The test failure seems dubious, could you try run it again?
The MSRV compilation error is caused by https://github.com/shadowsocks/shadowsocks-rust/pull/1992#discussion_r2232098179

src/config.rs
zonyitoo: You should keep the MSRV. Bumping MSRV should be a breaking change.

src/config.rs
zonyitoo: Yes indeed. max_files should always be >= 1.

src/config.rs
shadowsocks69420: Thanks for the feedback, I've updated the changes accordingly.

shadowsocks69420:
Closing this as the alternative approach #1993 has been merged.

Issue #1990, opened 2025-07-18
Consider an alterantive solution for logging to file when running as Windows service


shadowsocks69420:
When running as a Windows service, stdout is generally inaccessible. Currently, file logging relies on log4rs, which warns that it will be deprecated and removed. Please consider providing an alternative solution for file-based logging on Windows or guidance for users running the service on Windows.

It's acceptable to only make the `sswinservice` binary log to a file.

Side note: The log4rs example in README.md should be removed if log4rs support is going away, to minimize impact from future breakage.

zonyitoo:
Is it Ok to just redirect outputs to one file?

shadowsocks69420:
> Is it Ok to just redirect outputs to one file?

AFAIK, that is not possible for Windows services. Even if it is possible, it lacks other crucial file logging utilities like log rotation, retention, etc.

zonyitoo:
https://docs.rs/tracing-appender/0.2.3/tracing_appender/rolling/struct.RollingFileAppender.html

It is nice that `tracing` have already provided a library to write output to rolling files.

shadowsocks69420:
> https://docs.rs/tracing-appender/0.2.3/tracing_appender/rolling/struct.RollingFileAppender.html
> 
> It is nice that `tracing` have already provided a library to write output to rolling files.

`tracing-appender` would definitely help. However, it doesn't support configuration from environment variables like `tracing_subscriber`. What do you think is the best way to let the user configure it?

zonyitoo:
The appender created from `tracing-appender` would be set as a `Writer` of `tracing-subscriber`.

shadowsocks69420:
> The appender created from `tracing-appender` would be set as a `Writer` of `tracing-subscriber`.

The question was how are we going to expose the configuration options to the end user, e.g.: by changing/breaking the `log.*` section of the `config.json` file? As `tracing-appender` doesn't have runtime external configuring support built-in.

zonyitoo:
Yes. Users could set options in the `log` object.

shadowsocks69420:
> Yes. Users could set options in the `log` object.

Gotcha. I will try to come up with a PR for this.

zonyitoo:
The `log` object could be extended with keys like `layer`, `formatter`, `appender` … which could allow us to add more other new configurable features supported by `tracing`.

shadowsocks69420:
[...]

PR #1993, opened 2025-07-29
feat: file and advanced logging configuration for tracing

shadowsocks69420:
This is the second attempt of https://github.com/shadowsocks/shadowsocks-rust/pull/1992, as requested in https://github.com/shadowsocks/shadowsocks-rust/issues/1990#issuecomment-3122188623

Compare to the first attempt, this iteration enables more flexibility for logging configuration, supporting logging to stdout and multiple files simultaneously, which was a limitation in the previous version.

This new PR is intended as an alternative to the first iteration, feel free to close the other once decision is made.

---

I removed `SS*Config` structs as the amount of code duplication involved to deserialize the new `LogWriterConfig` enum is too much. And I don't really get the benefit of having them in the first place. I can see it makes sense for server configurations for backward compatibility reasons, but service configurations don't carry the same baggage and, in most cases, can be easily worked around by utilizing existing `serde_derive` features.

I've provided extra tests to ensure the change to the deserialization logic won't break existing configurations.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -5,7 +5,6 @@ use std::{
     fs::OpenOptions,
     io::{self, Read},
     path::{Path, PathBuf},
-    str::FromStr,
 };
 
 use clap::ArgMatches;
@@ -96,7 +95,8 @@ pub enum ConfigError {
 }
 
 /// Configuration Options for shadowsocks service runnables
-#[derive(Debug, Clone, Default)]
+#[derive(Deserialize, Debug, Clone, Default)]
+#[serde(default)]
 pub struct Config {
     /// Logger configuration
     #[cfg(feature = "logging")]
@@ -120,54 +120,7 @@ impl Config {
 
     /// Load `Config` from string
     pub fn load_from_str(s: &str) -> Result<Self, ConfigError> {
-        let ssconfig = json5::from_str(s)?;
-        Self::load_from_ssconfig(ssconfig)
-    }
-
-    fn load_from_ssconfig(ssconfig: SSConfig) -> Result<Self, ConfigError> {
-        let mut config = Self::default();
-
-        #[cfg(feature = "logging")]
-        if let Some(log) = ssconfig.log {
-            let mut nlog = LogConfig::default();
-            if let Some(level) = log.level {
-                nlog.level = level;
-            }
-
-            if let Some(format) = log.format {
-                let mut nformat = LogFormatConfig::default();
-                if let Some(without_time) = format.without_time {
-                    nformat.without_time = without_time;
-                }
-                nlog.format = nformat;
-            }
-
-            if let Some(config_path) = log.config_path {
-                nlog.config_path = Some(PathBuf::from(config_path));
-            }
-
-            config.log = nlog;
-        }
-
-        if let Some(runtime) = ssconfig.runtime {
-            let mut nruntime = RuntimeConfig::default();
-
-            #[cfg(feature = "multi-threaded")]
-            if let Some(worker_count) = runtime.worker_count {
-                nruntime.worker_count = Some(worker_count);
-            }
-
-            if let Some(mode) = runtime.mode {
-                match mode.parse::<RuntimeMode>() {
-                    Ok(m) => nruntime.mode = m,
-                    Err(..) => return Err(ConfigError::InvalidValue(mode)),
-                }
-            }
-
-            config.runtime = nruntime;
-        }
-
-        Ok(config)
+        json5::from_str(s).map_err(ConfigError::from)
     }
 
     /// Set by command line options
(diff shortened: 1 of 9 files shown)


src/config.rs
```diff
@@ -198,31 +151,127 @@ impl Config {
...
-#[derive(Debug, Clone, Default)]
+#[derive(Deserialize, Debug, Clone)]
+#[serde(default)]
 pub struct LogConfig {
-    /// Default logger log level, [0, 3]
+    /// Default log level for all writers, [0, 3]
     pub level: u32,
-    /// Default logger format configuration
+    /// Default format configuration for all writers
     pub format: LogFormatConfig,
```
shadowsocks69420, line 167: My original plan was to remove these fields from the struct while keeping the configuration format backward-compatible. E.g.: create a stdout writer using these values if no `writers` is configured, but ignore them otherwise (because it should be configured on a per-writer basis).

However, having to support the command line option prevented me from doing so because I figured that the command line option has to be handled in a somewhat sensible way and not ignored. But it is hard to reason about when there can be multiple writers. E.g.: which writer should the command line option apply to?

I'd still like to remove them if possible, let me know if you have opinion on this,

shadowsocks69420 (commented):
Pointing out some changes that are worth discussing

src/config.rs
```diff
@@ -198,31 +151,127 @@ impl Config {
...
-    /// Logging configuration file path
+    /// Log writers configuration
+    pub writers: Vec<LogWriterConfig>,
```
shadowsocks69420, line 169: I used the term `writers` as requested in your comment https://github.com/shadowsocks/shadowsocks-rust/issues/1990#issuecomment-3122188623

However, please note that in `tracing` the term `writer` refers the `MakeWriter` trait which makes `io::Write`. What each `LogWriterConfig` here corresponds to is a "subscriber" or "layer" in tracing-speak.

shadowsocks69420:
@zonyitoo This is ready for review. Waiting for your feedback on this.

zonyitoo:
I know. But I was quite busy these days.

robot-dot-win:
My current log.yaml is:
```yaml
appenders:
    logfile:
        kind: rolling_file
        #path: c:\\ss\\log\\ssl.log  # Windows
        path: /var/log/ss/ssl.log
        encoder:
            kind: pattern
            pattern: "{d(%Y.%m.%d %H:%M:%S)} {l} {M} - {m}{n}"
        policy:
            kind: compound
            trigger:
                kind: size
                limit: 10 mb
            roller:
                kind: fixed_window
                #pattern: "c:\\ss\\log\\ssl.log.{}"   # Windows
                pattern: "/var/log/ss/ssl.log.{}"
                base: 1
                count: 20

root:
    level: info
    appenders:
        - logfile

loggers:
    shadowsocks_service::local::utils:
        level: debug
    shadowsocks_service::local::socks::server:
        level: debug
```
I'm a user, not a go programmer. As for the new logger, how to config different service to use different log level? Thanks.

zonyitoo:
1. This is a Rust project, not Go.
2. Logging configuration applies globally. What do you mean "services"?

robot-dot-win:
> 1. This is a Rust project, not Go.
> 2. Logging configuration applies globally. What do you mean "services"?

1. Sorry for my mistaking shadowsockes for coredns :)
2. I meant that I would set the root log level to info, but set shadowsocks_service::local::utils to debug, just like what I'm using via log.yaml.

shadowsocks69420:
@robot-dot-win The new logging implementation will still honer the `RUST_LOG` environment variable. The syntax is documented [here](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#example-syntax). Though keep in mind that the environment variable applies/overrides to all writers globally.

robot-dot-win:
> @robot-dot-win The new logging implementation will still honer the `RUST_LOG` environment variable. The syntax is documented [here](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#example-syntax). Though keep in mind that the environment variable applies/overrides to all writers globally.

Thanks. My linux systemd service config:
```ini
[Unit]
Description=Shadowsocks Client
Documentation=https://github.com/shadowsocks/shadowsocks-rust
After=network.target

[Service]
Type=simple
User=root
Environment="RUST_LOG=info,shadowsocks_service::local::utils=debug,shadowsocks_service::local::socks::server=debug"
ExecStart=/opt/ss/sslocal --config /etc/ss/ssl.json
Restart=on-failure

[Install]
WantedBy=multi-user.target
```
And the log config for sslocal:
```json
    "log": {
        "level": 2,
        "format": {
            "without_time": false,
        },
        "writers": [
            {
                "file": {
                    "directory": "/var/log/ss",
                    "rotation": "daily",
                    "max_files": 32
                }
            }
        ]
    }
```
```
[root@ss]# systemctl status ssl
● ssl.service - Shadowsocks Client
     Loaded: loaded (/usr/lib/systemd/system/ssl.service; enabled; preset: disabled)
     Active: active (running) since Sat 2025-09-06 14:06:52 CST; 18min ago
       Docs: https://github.com/shadowsocks/shadowsocks-rust
   Main PID: 5497 (sslocal)
      Tasks: 12 (limit: 48880)
     Memory: 2.1M (peak: 3.4M)
        CPU: 13ms
     CGroup: /system.slice/ssl.service
             └─5497 /opt/ss/sslocal --config /etc/ss/ssl.json

Sep 06 14:06:52 ss systemd[1]: Started Shadowsocks Client.
Sep 06 14:06:52 ss sslocal[5497]: 2025-09-06T14:06:52.33398795+08:00  INFO main ThreadId(01) shadowsocks_rust::service::local: shadowsocks local 1.23.5 build 2025-07-04T15:22:50.619501999+00:00
Sep 06 14:06:52 ss sslocal[5497]: 2025-09-06T14:06:52.336667036+08:00  INFO tokio-runtime-worker ThreadId(10) shadowsocks_service::local::socks::server::server: shadowsocks socks TCP listening on 192.168.252.48:1080
```
The service runs successfully, but there's nothing created in the directory /var/log/ss/.

Anything wrong with my config?

shadowsocks69420:
@robot-dot-win I see that you are using the 1.23.5 version built on July 4. However, the new implementation in this PR was merged on Aug 29 and it hasn't been released yet. If you want to try out the new feature before a release, you need to build the binary from `master` yourself.

robot-dot-win:
> @robot-dot-win I see that you are using the 1.23.5 version built on July 4. However, the new implementation in this PR was merged on Aug 29 and it hasn't been released yet. If you want to try out the new feature before a release, you need to build the binary from `master` yourself.

Thanks for your comment. I thought it had been released when I checked the Configuration section of README. Better to keep README consistent with the latest release.

robot-dot-win:
BTW, I'm not a rust programmer, but does the tracing_subscriber have to depend on the environment variable RUST_LOG? You can set environment variables in a Linux service config, but Windows services do not support that, and you have to write a powershell script to wrap the application.

shadowsocks69420:
@robot-dot-win 

> but does the tracing_subscriber have to depend on the environment variable RUST_LOG?

That is a convention used by logging libraries in Rust ecosystem, not just `tracing_subscriber`. See [`env_logger`] for example.

[`env_logger`]: https://docs.rs/env_logger/latest/env_logger/#enabling-logging

> but Windows services do not support that, and you have to write a powershell script to wrap the application.

It is possible to set environment variable for a Window service, though a bit more involved:
https://serverfault.com/questions/813506/setting-environment-variable-for-service/1103091#1103091

robot-dot-win:
@shadowsocks69420 Thank you so much, all cleared up!

PR #1995, opened 2025-07-30
feat: add selinux module

f100024:
This policy provides several security improvements over running shadowsocks as `unconfined_service_t`:

- **Principle of least privilege**: Only grants necessary permissions
- **Network isolation**: Controls which ports and connections are allowed
- **File system protection**: Restricts file access to configuration and required system files
- **Process isolation**: Runs in a dedicated SELinux domain
- **Audit trail**: All access attempts are logged for security monitoring

### DIFF
--- a/selinux/README.md
+++ b/selinux/README.md
@@ -1,0 +1,106 @@
+# Shadowsocks SELinux Policy
+
+## Prerequisites
+
+Install required SELinux development tools:
+```bash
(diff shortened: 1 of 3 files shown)


zonyitoo:
In the security points of view, how is SELinux compared with Docker? I think the Docker or cgroup related solution could provide more restriction about isolation.

PR #1996, opened 2025-07-30
fix(docker): replace musl.cc

AaronChen0:
Tested on https://github.com/AaronChen0/shadowsocks-rust/actions/runs/16624737890 with https://github.com/OpenListTeam/musl-compilers/releases/download/2025-06-12/$MUSL-cross.tgz.
But later I decided to make a mirror repo for my own. I will update the links if musl.cc get any updates.

### DIFF
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -3,6 +3,7 @@ on:
   push:
     tags:
       - v*
+  workflow_dispatch:
 
 jobs:
   build-docker-image:
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,7 +27,7 @@ RUN case "$TARGETARCH" in \
         exit 1 \
     ;; \
     esac \
-    && wget -qO- "https://musl.cc/$MUSL-cross.tgz" | tar -xzC /root/ \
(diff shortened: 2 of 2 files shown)


shadowsocks69420:
I think sha512sum check at build time should be added for this one as the content of the link it points to can be altered by the author at anytime without anyone noticing.

@zonyitoo For the time being, I suggest we create a new repository under the [shadowsocks org](https://github.com/shadowsocks) then move the same artifacts to the newly created repository so that only trusted individuals can modify them. This should be fairly trivial IMO.

AaronChen0:
I had actually thought about this issue before, but I am not familiar with docker at all, never use it. Pull requests are welcome if you can add hardcoded sha512sum check.

Yes. Moving the artifacts to a new repo of [shadowsocks org](https://github.com/shadowsocks) is preferred.

zonyitoo:
> I think sha512sum check at build time should be added for this one as the content of the link it points to can be altered by the author at anytime without anyone noticing.
> 
> @zonyitoo For the time being, I suggest we create a new repository under the [shadowsocks org](https://github.com/shadowsocks) then move the same artifacts to the newly created repository so that only trusted individuals can modify them. This should be fairly trivial IMO.

That is reasonable. @madeye what do you think?

PR #2029, opened 2025-10-11
fix: fake ips should not be passed to ACL rules

### LINKED ISSUE #2028: ACL checks should based on associated domain when `fake-dns` is enabled
According to the code snippet here:
https://github.com/shadowsocks/shadowsocks-rust/blob/2fe2b07ae28197916b273da02bb8656b6fecbd5d/crates/shadowsocks-service/src/local/net/tcp/auto_proxy_stream.rs#L52-L57
https://github.com/shadowsocks/shadowsocks-rust/blob/2fe2b07ae28197916b273da02bb8656b6fecbd5d/crates/shadowsocks-service/src/local/net/tcp/auto_proxy_stream.rs#L69-L85

The current implementation checks the ACL **before** mapping fake-dns IPs to their associated domain. This results the ACL is always acting on fake IPs, which is likely not what user would expect and has no meaningful use case in my opinion.

As the title suggests, I believe we should adjust the implementation so that the ACL decisions will be based on the associated domain instead of allocated fake IPs.

shadowsocks69420:
ACL rules are likely not written for fake IPs.

One of the major selling point of using the `fake-dns` feature is to be able to make use of ACL rules that are based on domain names instead of purely IP addresses. Passing fake IPs to ACL nullifies this benefit, which is likely not expected from users.

Closes #2028

### DIFF
--- a/crates/shadowsocks-service/src/local/net/tcp/auto_proxy_stream.rs
+++ b/crates/shadowsocks-service/src/local/net/tcp/auto_proxy_stream.rs
@@ -8,6 +8,7 @@ use std::{
     task::{self, Poll},
 };
 
+use log::trace;
 use pin_project::pin_project;
 use shadowsocks::{
     net::{ConnectOpts, TcpStream},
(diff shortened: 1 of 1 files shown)


shadowsocks69420:
Because I'm lacking thorough understanding of the codebase, I wasn't confident to remove the same mapping logic in the functions that immediately get called, as I'm worried that there might be other direct calls to these functions that expects it to map the fake-dns IPs.

https://github.com/shadowsocks/shadowsocks-rust/blob/16ff8c8bc90adbba4f0d9dc53c12f81e2d25abf1/crates/shadowsocks-service/src/local/net/tcp/auto_proxy_stream.rs#L68-L86

https://github.com/shadowsocks/shadowsocks-rust/blob/16ff8c8bc90adbba4f0d9dc53c12f81e2d25abf1/crates/shadowsocks-service/src/local/net/tcp/auto_proxy_stream.rs#L96-L111

Also, I did not remove the trace log as I think might be helpful for debugging purposes, I couldn't find other trace logs that provide the ACL decision.

zonyitoo:
Why not make a private function like `connect_bypassed_with_opts_impl` to prevent mapping fake ips twice.

Issue #1875, opened 2025-02-16
no ssservice in snap install?


jkldgoefgkljefogeg:
```
$ snap info shadowsocks-rust
name:      shadowsocks-rust
summary:   Rust port of Shadowsocks
publisher: DuckSoft
store-url: https://snapcraft.io/shadowsocks-rust
license:   MIT
description: |
  Shadowsocks Client & Server, written in Rust.
commands:
  - shadowsocks-rust.sslocal
  - shadowsocks-rust.ssmanager
  - shadowsocks-rust.ssserver
  - shadowsocks-rust.ssurl
services:
  shadowsocks-rust.sslocal-daemon:  simple, disabled, inactive
  shadowsocks-rust.ssserver-daemon: simple, disabled, inactive
snap-id:      L7MfxJ3tx3xxgeKsu5ouPiaWwiWE56yx
tracking:     latest/stable
refresh-date: today at 19:05 CST
channels:
  latest/stable:    v1.22.0.r0.g7e059ca6  2025-01-02 (1964) 15MB -
  latest/candidate: v1.22.0.r0.g7e059ca6  2025-01-02 (1964) 15MB -
  latest/beta:      v1.22.0.r0.g7e059ca6  2025-01-02 (1964) 15MB -
  latest/edge:      v1.22.0.r40.g8b56aa2c 2025-02-15 (2004) 15MB -
installed:          v1.22.0.r0.g7e059ca6             (1964) 15MB -

ls /snap/bin/shadowsocks*
/snap/bin/shadowsocks-rust.sslocal    /snap/bin/shadowsocks-rust.ssserver
/snap/bin/shadowsocks-rust.ssmanager  /snap/bin/shadowsocks-rust.ssurl

```

zonyitoo:
Currently yes. You could make a PR if you want to add a new command.

PR #2039, opened 2025-11-01
allow snap to install ssservice script

leeylf:


zonyitoo:
Why would you need that in snap?

leeylf:
> Why would you need that in snap?

I was trying to follow the instructions to setup a server and realize I can't run `ssservice keygen`. I also saw a similar complain [here](https://github.com/shadowsocks/shadowsocks-rust/issues/1875) where you actually suggested to make a PR.

Issue #1397, opened 2023-12-31
[Feature Request]: Allow to chain ssserver to another socks5 proxy


vt-alt:
Please allow to chain shadowsocks server to another socks proxy.

For example user wants to connect to a cloudflare Warp gateway (which is in socks5 proxy mode) but to have his side using shadowsocks (on windows).
Currently shadowsocks-windows do not allow to configure `tunnel` protocol. This prevents such scheme.

So it would allow to use usual shadowsocks windows client (as a socks proxy) that connects as normal to a remote ssserver but that ssserver _redirects_ (not decoded socks5 traffic) to a local warp-server socks5 port.

zonyitoo:
I can’t get your point. Could you elaborate it more in detail?

vt-alt:
Idea is that `ssserver` would not terminate connections on its own but to forward them to some other socks server (such as Warp in my example).

Or from another point of view it's to allow to configure `tunnel` mode on the `ssserver` (while now it's only for `sslocal`).

zonyitoo:
> Or from another point of view it's to allow to configure `tunnel` mode on the `ssserver` (while now it's only for `sslocal`).

Of course. `ssserver` is designed as a remote server of shadowsocks, so it is only for working with `sslocal`.

> Idea is that `ssserver` would not terminate connections on its own but to forward them to some other socks server (such as Warp in my example).

`ssserver` is running in a remote server. It's outbound is to connect target remote servers. Why were targets are connected via a socks5 server?

vt-alt:
In my example browser connects with socks5 to a local shadowsocks-windows, but shadowsocks-windows does not allow to configure `tunnel` protocol. So idea is to allow remote ssserver (which local shadowsocks connects to) to also allow configuring tunnel or chaining mode. 

(ps. This may also be useful for general proxy chaining, like terminating all ssserver connections to Tor which runs on the same remote server as ssserver.)

zonyitoo:
If I understand it correctly: your `ssserver` was hidden from `sslocal` with a socks5 proxy.

```
Client -> sslocal -----(socks5)-----> ssserver -> Target
```

So the key issue in this topology is that `sslocal` must connect to `ssserver` with socks5 proxy protocol.

It should be Ok to add Proxy support in the `ServerConfig`, for example:

```json
{
	"servers": [
		{
			"server": "1.2.3.4",
			"server_port": 8999,
			"proxy": "socks5://2.3.4.5:1080"
		}
[...]

PR #2106, opened 2026-04-09
[Feature]: Allow to chain ssserver to another socks5/http/https proxy #1397

missdeer:
Resolve issue #1397

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3331,6 +3331,7 @@ dependencies = [
  "tokio-rustls",
  "trait-variant",
  "tun",
+ "url",
  "webpki-roots 1.0.6",
  "windows-sys 0.61.2",
  "zstd",
--- a/README.md
+++ b/README.md
@@ -372,6 +372,24 @@ Start local client with configuration file
 sslocal -c /path/to/shadowsocks.json
 ```
 
+`sslocal` also supports routing its outbound TCP connection to the Shadowsocks server through a proxy or proxy chain with the `outbound_proxy` config key. Supported hop types are `socks5://`, `http://`, and `https://`, with optional `user:pass@` credentials. This option is currently available through the configuration file for `sslocal`; `ssserver` supports both the config file and repeated `--outbound-proxy` command line flags.
+
+```jsonc
+{
+    "server": "server.example.com",
+    "server_port": 8388,
+    "password": "hello-kitty",
+    "method": "aes-256-gcm",
+    "local_address": "127.0.0.1",
+    "local_port": 1080,
+    "outbound_proxy": [
+        "socks5://user:pass@127.0.0.1:1080",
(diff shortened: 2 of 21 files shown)


zonyitoo:
It looks like AI generated codes. :)

The HTTPS proxy implementation could just use `shadowsocks-service/src/local/http/http_client.rs`, which implements a standard HTTP client with HTTPS support.

`outbound_proxies` could also be supported on the local side, check the client stream in `local/net/tcp/auto_proxy_stream.rs`, `local/net/association.rs`.

zonyitoo:
The core design is Ok to me, just solve the compliation errors and the HTTPS design issue.

missdeer:
Updated this branch with the latest local changes.
    - `sslocal`: the TCP connection from local mode to the Shadowsocks server can also go through the same proxy chain.
  - The chain is TCP-only. UDP is not proxied by `outbound_proxy`.
  - `ssserver` supports both JSON config and repeated `--outbound-proxy` CLI flags.
  - `sslocal` currently supports `outbound_proxy` through JSON config; there is no CLI flag yet.

  HTTPS design:

  - I did not route the chain implementation through the full Hyper-based `local/http/http_client.rs` request path.
  - Instead, I extracted the HTTPS transport wrapper into a shared low-level helper in `shadowsocks-service/src/net/http_stream.rs`.
  - Both the local HTTP client and the outbound proxy-chain code now use that same TLS transport helper, so the TLS setup is shared while the actual HTTP `CONNECT` tunnel establishment
  remains a small low-level helper.
  - This keeps the proxy-chain code stream-oriented, which fits the need here better than forcing it through a request/response client abstraction.

  Runtime fixes included in this update:

  - fixed local-side SOCKS5 success replies for chain-backed streams by preserving the underlying local socket address
  - fixed chained HTTPS proxy hop ordering so TLS is established before talking to each HTTPS hop
  - fixed `ssserver` so the first outbound proxy hop still honors existing outbound `ConnectOpts` / resolver behavior instead of bypassing them with raw `TcpStream::connect`

  Local verification:

  - `cargo check --workspace`
  - `cargo check -p shadowsocks-service --features "full local-http-native-tls"`
  - `cargo test -p shadowsocks-service outbound_proxy --lib`

  Docs were also updated in README to match the actual config / CLI surface.

Aleximal:
> Updated this branch with the latest local changes. - `sslocal`: the TCP connection from local mode to the Shadowsocks server can also go through the same proxy chain.
> 
> * The chain is TCP-only. UDP is not proxied by `outbound_proxy`.
> * `ssserver` supports both JSON config and repeated `--outbound-proxy` CLI flags.
> * `sslocal` currently supports `outbound_proxy` through JSON config; there is no CLI flag yet.


Awwww no love for UDP? Sad, I had such plans for this 😅😭

PR #2111, opened 2026-04-16
Remove unused imports in outbound_proxy

faern:
Just a tiny and simple cleanup of unused imports, causing compiler warnings


faern:
I also upgraded two dependencies that have advisories posted against them. Otherwise the CI failed. I did not break these out to their own PRs, because then I suspected that the import issues would block *that* PR, and just creating a catch 22 :)

PR #2112, opened 2026-04-16
Replace `ring` with `aws-lc-rs`

faern:
Ring is unmaintained. Even if there are no known vulnerabilities in it it's a risk to depend on it. Since it's unmaintained it probably gets fewer eyes on it studying if it's correctly implemented. If something bad is found in `ring` it's going to be harder to patch compared to if we use a more maintained crypto implementation.

`aws-lc-rs` is more maintained and seems to be what most move towards. `aws-lc-rs` expose more cryptographic primitives, that `ring` does not have. Even if `shadowsocks` itself does not need these newer things, downstream users of `shadowsocks` might (we are one of those!) and they probably don't want to have *both* ring and aws-lc-rs in their dependency trees as it blows up both compile time and binary size.

Luckily it seems pretty easy to migrate all of shadowsocks to aws-lc-rs, as you can see in this PR. It just depends on https://github.com/shadowsocks/shadowsocks-crypto/pull/24 being merged and published first.

I have not tested this on Windows myself. I hope the CI is good enough.

This PR builds on top of #2111. Because I wanted the CI to be as clean as possible in order to catch any issues this PR might introduce. So please merge that cleanup PR first, then I can rebase this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -240,6 +240,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a054912289d18629dc78375ba2c3726a3afe3ff71b4edba9dedfca0e3446d1fc"
 dependencies = [
  "aws-lc-sys",
+ "untrusted 0.7.1",
  "zeroize",
 ]
 
@@ -260,12 +261,6 @@ name = "base64"
 version = "0.22.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
-
-[[package]]
-name = "base64ct"
-version = "1.8.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7d809780667f4410e7c41b07f52439b94d2bdf8528eeedc287fa38d3b7f95d82"
 
 [[package]]
 name = "bindgen"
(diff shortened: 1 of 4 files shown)


faern:
I believe the only thing left here before marking the PR as ready for review, is a published release of `shadowsocks-crypto`, so we can get rid of the git dependency.

Or do you want the `shadowsocks` crate to also have selectable crypto backend? That is going to complicate things a bit, but is maybe doable. What's important to me as a consumer of `shadowsocks` as a *library* is that I can depend on it without having `ring` pulled into my dependency tree, that's it basically.

zonyitoo:
Since I didn't make `ring` a configurable option in `shadowsocks` crate, I think it should be Ok to make `aws-lc` to be the default crypto backend for `shadowsocks`. To get rid of `ring` completely, we should do the same thing on the `rustls` features list.

I am still working on refactoring the project. After that please update this PR with the latest master.

faern:
Yes, the ring features on `rustls` have already been addressed in this PR. With this PR applied, `shadowsocks` does not depend on ring.

Good luck with the refactoring. I hope you are not aiming to do too big stuff, that might stall a ring-free release :pray:. I will rebase this when I see updates.

faern:
Updated with the `shadowsocks-crypto` 0.7.0 release. So this should now be ready for review! The only error in CI is about the advisory against `hickory-resolver` which is not caused by this PR but simply by that advisory recently being published.

faern:
@zonyitoo Can we get a release with this change? Or do you have a plan for what will be included/when the next release will come out?

zonyitoo:
I am waiting for those crypto libraries comes from RustCrypto to upgrade to their latest version.
https://github.com/shadowsocks/shadowsocks-crypto/pull/25

PR #2161, opened 2026-07-26
fix: allow fragmentation for inbound UDP replies

pengwk:
## Summary

- honor the existing `outbound_udp_allow_fragmentation` option on inbound UDP listener sockets
- propagate global and per-server settings, including manager mode
- preserve the current default and cover Unix and Windows
- add a Linux regression test for `IP_MTU_DISCOVER`

## Problem

The UDP relay sends encrypted responses back to clients through its inbound listener socket. That socket always enabled DF, even when `outbound_udp_allow_fragmentation` was set. A near-MTU target response can grow after encryption and then fail with `EMSGSIZE` on the client-facing send path.

This change reuses the existing option instead of adding another configuration knob, making it apply to both UDP relay directions.

## Verification

- Linux: `cargo test -p shadowsocks --test udp --no-default-features --features aead-cipher-2022` (4 passed)
- Linux: `cargo check -p shadowsocks-service --no-default-features --features server,aead-cipher-2022`
- Real relay path: a 1472-byte UDP response failed before the change and passed 10/10 after it with fragmentation enabled

### DIFF
--- a/README.md
+++ b/README.md
@@ -842,6 +842,8 @@ Example configuration:
             "outbound_bind_addr": "11.22.33.44",
             // Outbound UDP socket allows IP fragmentation (default false)
             "outbound_udp_allow_fragmentation": false,
+            // Inbound UDP socket allows IP fragmentation (default false)
+            "inbound_udp_allow_fragmentation": false,
             // Route outbound TCP connections through a proxy or proxy chain
             // (TCP only; UDP is not proxied)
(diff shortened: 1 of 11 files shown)


zonyitoo:
The `outbound_udp_allow_fragmentation` flag is designed for "outbound" sockets. So I added a new `inbound_udp_allow_fragmentation`.

